PR #14, opened 2018-05-30
feat(npm): updated scripts and related doc

mminot-yseop:
1. It seems perfectly useless to make sure the working directory is the same at the end of an npm script and at the beginning.
2. I tend to use subshells spawned through parentheses in order to restore the previous working directory instead of using `cd ..`, etc. Among other benefits, it makes it useless to use weird conditions like `&&` when trying to guess if we succeeded in going to a directory or not. Basically, I went from `cd a && cmd && cd ..` to `cd a && { cmd; cd .. }` and finally to `(cd a && cmd)`. The first form, which was used before my edits, is not really robust as it leaves the working directory changed to `a` if `cmd` fails.
3. Updated doc to tell people to install the `vsce` module locally. This seems to help `npm` to find it when running the `package` script.
4. More info on what to do if this still fails.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -39,25 +39,6 @@ The format available in this manual is a little bit different than the one we ar
 
 ## Package the extension
 
-Note that the commands shown here should preferably be executed while being in the vscode-yseopml project's directory, except when stated otherwise.
-
-### Prerequisites
-
-Make sure you have installed the `vsce` node package in the project's modules:
-
-```[bash]
-npm install vsce
-```
-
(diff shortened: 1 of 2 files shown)


mminot-yseop:
By the way, maybe `vsce` should be added in `devDependencies` or something. Dunno. Might be nice to install it automatically in the project's modules.

mminot-yseop:
Here! Now you can brutally remove the `node_modules` directory of the project, run `npm run package`, and the `vsce` module should be downloaded automatically and correctly found during the execution of the script.

PR #18, opened 2018-05-31
doc(snippets): add list and, in a comment, a command to update it

mminot-yseop:
Self-explanatory and trivial.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -70,3 +70,16 @@ code --install-extension EXTENSION_FILE.vsix
 . Select the extension file and validate.
 . Reload {vsc}.
 . Open `client/test/test.kao` to check that the extension is working properly.
+
+
+## Contributing
+
+You're free to contribute to this project as long as you respect its guidelines.
+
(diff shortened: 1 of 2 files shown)


alacambre-yseop:
I think the comments should be in the root README file. Also, there is nothing explaining what is jq nor how to get it.

mminot-yseop:
I dunno… The root readme seems less contributors-oriented, to me. I wanted to intercept anyone who tried to change the client readme after having added a feature, so…
As for `jq`, they can look it up on the web, but I'll see what I can do…

mminot-yseop:
Seriously, there's nothing even remotely related to features, snippets or contributions in the global readme.

alacambre-yseop:
Well, there are a bunch of informations about antlr4 or syntax highlighting stuff. There are even informations about packaging. This is clearly contributors-oriented.

mminot-yseop:
Still, a command meant to generate stuff for a given file should probably be show in the very same file.

I'm planning on adding a “Contributing” section to the global readme, but this brings new questions to the surface: for example: should the contributors update the client-readme themselves in their PRs, or should we do it ourselves during the releasing process?

mminot-yseop:
Here…

PR #19, opened 2018-06-10
feat(model): load yseop engine model + provide YE types completion

alacambre-yseop:


client/package.json
```diff
@@ -42,6 +42,12 @@
                     "default": 100,
                     "description": "Controls the maximum number of problems produced by the server."
                 },
+                "yseopml.pathToPredefinedObjectsYml": {
+                    "scope": "window",
+                    "type": "string",
+                    "default": "",
+                    "description": "Full path to an xml file containing the YML types definition."
```
mminot-yseop: “xml” ↔ “YML”
Please be more consistent regarding acronyms. I would advocate using uppercase letters for all “—ML” stuff.
Oh, and as I said somewhere else, use “absolute” or “relative path” rather than “full path”. I never really know what you mean by that.

client/server/EngineModel.js
```diff
@@ -0,0 +1,72 @@
...
+var xml2js = require('xml2js');
+var parser = new xml2js.Parser();
+class EngineModel {
+    constructor(path, completionItems) {
+        this.path = path;
+        this.completionItems = completionItems;
+        this.loadPredefinedObjects();
+    }
+    loadPredefinedObjects() {
+        // load predefined objects
```
mminot-yseop: I laughed at that comment. It's just a transcription of the function's name.

client/server/EngineModel.js
```diff
@@ -0,0 +1,72 @@
...
+        this.path = path;
+        this.completionItems = completionItems;
+        this.loadPredefinedObjects();
+    }
+    parsePredefinedObjects(path, completionItems) {
+        console.log(`Parsing definition file: ${path}`);
+        fs.readFile(path, function (err, data) {
+            parser.parseString(data, function (err, result) {
+                if (err != null || result == null) {
+                    console.error("Something went wrong during YE model import:\n" + err);
```
mminot-yseop: Since the condition is `err != null || result == null`, we might end up logging a `null` error, right?

client/server/EngineModel.js
```diff
@@ -0,0 +1,72 @@
...
+        this.completionItems = completionItems;
+        this.loadPredefinedObjects();
+    }
+    loadPredefinedObjects() {
+        // load predefined objects
+        if (fs.existsSync(this.path)) {
+            this.parsePredefinedObjects(this.path, this.completionItems);
+        }
+        else {
+            console.warn(`File “${this.path}” doesn't exist.`);
```
mminot-yseop: Shouldn't this be an error rather than a warning? I guess lots of things will go wrong afterwards.

client/server/package.json
```diff
@@ -18,13 +18,14 @@
         "@types/node": "^6.0.88",
         "antlr4ts": "^0.4.1-alpha.0",
         "antlr4ts-cli": "^0.4.0-alpha.4",
-        "typescript": "^2.8.4",
```
mminot-yseop: ? The required version for TS went down?

client/server/server.js
```diff
@@ -40,13 +41,22 @@ connection.onInitialize((_params) => {
...
+let pathToPredefinedObjectsYml;
 // The settings have changed. Is send on server activation
 // as well.
 connection.onDidChangeConfiguration((change) => {
     let settings = change.settings;
     maxNumberOfProblems = settings.yseopml.maxNumberOfProblems || 100;
+    pathToPredefinedObjectsYml = settings.yseopml.pathToPredefinedObjectsYml;
+    if (engineModel == null) {
+        engineModel = new EngineModel_1.EngineModel(pathToPredefinedObjectsYml, completionItems);
+    }
```
mminot-yseop: You've got quite a lot of

```
}
else {
```

in your JS code. That's surprising.

client/server/server.js
```diff
@@ -40,13 +41,22 @@ connection.onInitialize((_params) => {
 documents.onDidChangeContent((change) => {
```
mminot-yseop: All these methods combining “on”, the “to do” verb and “change” (another verb) make it sound as if it was a “did” / “do” use to put an emphasis over something that _did_ actually happen. Weird. Why not just `onChange(d)—`?

package.json
```diff
@@ -30,4 +30,4 @@
         "typescript": "^2.5.2",
         "vsce": "^1.41.0"
     }
-}
\ No newline at end of file
+}
```
mminot-yseop: OK… xD You keep removing newlines in some files and adding them back in others. Your editor might be doing weird things.

server/src/EngineModel.ts
```diff
@@ -0,0 +1,82 @@
+import {
```
mminot-yseop: Woops, I might have commented on the compiled version. Well, go back there to see what I wrote, I guess… xD

mminot-yseop:
Build OK. Dunno what I'm supposed to do to test.

client/server/server.js
alacambre-yseop: Ask Microsoft :smile:

client/server/server.js
alacambre-yseop: The JS is generated from the `*.ts` files.

client/server/EngineModel.js
alacambre-yseop: Nothing will go wrong. We just ignore the parsing of the file. However, maybe we should warn user directly with a notification? The path is added directly by the user, after all.

alacambre-yseop:
@mminot-yseop I added a new parameter wich will contain the absolute path to a `predefinedObjects.xml` file.
The server-side code will try to parse this file and add as suggestions all of the Yseop Engine type names.

mminot-yseop:
So I must set this new parameter myself, right?

alacambre-yseop:
Yes :smile:

mminot-yseop:
I'm blindly adding `"yseopml.pathToPredefinedObjectsYml": "/home/mminot/yseop-engine/yseop-engine-7.0.0.RC05/ymlPredefinedObjects.xml"` to my preferences. No doc on what the user is actually supposed to do and how? (and there might be confusion between `predefinedObjects.xml` and `ymlPredefinedObjects.xml`.

mminot-yseop:
I can't seem to be able to get any result. o( >_< )o

alacambre-yseop:
I think, maybe we should let user just set his path to the YE she wanna use, and add ourseleves the expectec filename?
The correct file is `predefinedObjects.xml`.

mminot-yseop:
C'mon maaaaan! Your property is called `pathToPredefinedObjectsYml` and you want the file _without_ “yml” in its name? xD

mminot-yseop:
But sure, the user should be able to set it as he wants, but some tips would be nice xD

mminot-yseop:
WOOO SHITLOADS OF TYPES POPPED UP

alacambre-yseop:
@mminot-yseop is that better?

client/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Release Notes
 
+## 1.X.X (yyyy-mm-dd)
+
+- Basic Yseop Engine model based type completion
```
mminot-yseop: Use an hyphen (“Yseop Engine model-based type completion”) or, since this isn't quite sufficient to make it pleasing to read, write something like “Basic type names completion, based on Yseop Engine's model”. If you agree.

client/README.md
```diff
@@ -11,8 +11,13 @@ This extension provides a Yseop Markup Language (YML) basic support for VS Code.
     * `granule`: Text granule
     * `instance`: Object instance
     * `textu`: Textualization attribute
+![Snippet for textualization Gif.](images/textualization.gif "Snippet for textualization.")
+
 - Code folding
 - KB class attributes completion (needs class to be opened at least once)
+- Basic Yseop Engine model based type completion
```
mminot-yseop: Copy whatever you write where I placed my first comment, I guess.

client/README.md
```diff
@@ -11,8 +11,13 @@ This extension provides a Yseop Markup Language (YML) basic support for VS Code.
...
+    * This feature uses the parameter `yseopml.pathToPredefinedObjectsXml` to get the **absolute** path of the `predefinedObjects.xml` file provided by user's version of Yseop Engine.
```
mminot-yseop: * No need to put so much stress on “absolute”. This will scare away the few ones that might not know what that means. Someone that wants to use a relative one will soon realize he has no idea what the starting directory should be, so…
* Path _to_ a file. “Of” makes it sound like it's an unique property or something.
* “user's version of Yseop Engine.” → grammatically weird. Plus, I guess the user can have several engine versions. Suggestion on the fly: “provided with one of the user's Yseop Engine versions”?

client/README.md
```diff
@@ -11,8 +11,13 @@ This extension provides a Yseop Markup Language (YML) basic support for VS Code.
     * `granule`: Text granule
     * `instance`: Object instance
     * `textu`: Textualization attribute
+![Snippet for textualization Gif.](images/textualization.gif "Snippet for textualization.")
```
mminot-yseop: Regarding the gif,
* I would have stopped before “number”, but the file's size if lower than I thought.
* Oh, and it's kinda weird that in the code you show there are three spaces before field names.
* And I would have removed about 150 (or even 250) pixels at the top and 180 at the bottom. And 300 pixels on the right. Don't forget this might end up being displayed in the “Extensions” panel of VSCode, which can be rather narrow (and cluttered) depending on what's opened alongside (and inside) it.

client/package.json
```diff
@@ -42,7 +42,7 @@
                     "default": 100,
                     "description": "Controls the maximum number of problems produced by the server."
                 },
-                "yseopml.pathToPredefinedObjectsYml": {
+                "yseopml.pathToPredefinedObjectsXml": {
```
mminot-yseop: Ah!

client/README.md
alacambre-yseop: * I wanted to show every options
* I don't “show” the spaces. I just click where I wanna start..
* Ok to remove pixels on top, but the right and the bottom's ones are usefull to display what happens in the suggestion list.

client/README.md
alacambre-yseop: Ok, I'll do the changes.

client/CHANGELOG.md
alacambre-yseop: Ok, I'll do the changes.

client/README.md
alacambre-yseop: Ok, I'll do the changes.

client/README.md
mminot-yseop: * Yeah but you could clean the surrounding code before doing anything (regarding spurious spaces).
* I know the suggestions pop up and take lots of space, but still, you kept quite a lot of it. The thing that goes the farther to the right is what is displayed right before you trigger the textualization snippet, and I based my horizontal calculations on that, so… 300 pixels. `*checking*` Well, 260, at least.
image

mminot-yseop: ImageMagick → `convert` or `mogrify` → `crop` + https://www.imagemagick.org/Magick++/Geometry.html

mminot-yseop: Omg omg

```
$ gifsicle --crop 0,250-895,690 --output new.gif textualization.gif
```

PR #41, opened 2018-10-31
feat(releaseScripts): basic script to produce a release branch

mminot-yseop:
Write date in changelog, update version number, package, commit, push.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,0 +1,72 @@
+#! /usr/bin/env bash
+
+set -e
+
+cd "$(dirname "$(readlink -f -- "$0")")"
+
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -0,0 +1,87 @@
...
+
+set -e
+
+cd "$(dirname "$(readlink -f -- "$0")")"
+
+declare -r deps=(
+    git
+    jq
+    npm
+    patate
```
alacambre-yseop: ???

mminot-yseop: :woman_facepalming:

mminot-yseop:
Sorry; should be better now.

PR #45, opened 2018-10-31
fix(releaseScripts): better way to tag releases

mminot-yseop:
This should ensure we do not place the tag on an old commit. Also tried to make things easier for the user.


mminot-yseop:
Squash after review

PR #58, opened 2019-01-04
test(server): add server-side tests 

alacambre-yseop:
PR to Squash before merge.


server/src/test/YmlToBdlVisitorImpl.test.ts
```diff
@@ -0,0 +1,143 @@
...
+import { YmlToBdlVisitorImpl } from "../YmlToBdlVisitorImpl";
+import { CompletionItem } from "vscode-languageserver";
+
+describe("Extension Server Tests", () => {
+  describe("YmlToBdlVisitorImpl", () => {
+    it("should be OK to instantiate YmlToBdlVisitorImpl", done => {
+      let inputStream = new ANTLRInputStream("");
+      let lexer = new YmlToBdlLexer(inputStream);
+      let tokenStream = new CommonTokenStream(lexer);
+      let parser = new YmlToBdlParser(tokenStream);
```
mminot-yseop: I was wondering whether using `const` around here for these things would be smart. Maybe it would make the flow slightly more evident. Dunno. TS noob.

mminot-yseop: (same kind of stuff further down in the diff)

server/tsconfig.test.json
```diff
@@ -0,0 +1,3 @@
+{
+	"extends": "./tsconfig.json"
```
mminot-yseop: Wait, does it still make sense to extend some config without overriding anything? I guess copying everything sure isn't an option, but… Well, I dunno. I guess this version is the best we can do? What if this file is deleted altogether?

alacambre-yseop: I think we should be able to delete this file safely.

PR #61, opened 2019-02-08
chore: upgrade dependencies

alacambre-yseop:


mminot-yseop:
The PR's name is a bit misleading. You appear to have fixed the style of 2572985739 files, too.

alacambre-yseop:
First, I have magic number phobia, second antlr files as well as all the package-lock.json files are automaticaly generated.

mminot-yseop:
It's not magic, it's a placeholder!

I was mostly surprised by the `' → "` thingies, but that does not matter much.

Will try to build and use that, I guess.

PR #63, opened 2019-02-13
chore: use tslint

alacambre-yseop:


mminot-yseop:
Did you apply a two-space indentation width on purpose? :thinking:

alacambre-yseop:
@mminot-yseop Yes, this is one of the rules of `"tslint:recommended"`

mminot-yseop:
Do you like it? I personally find that hard to read.

client/src/extension.ts
```diff
@@ -2,188 +2,212 @@
...
+function updateSettings(response: string): void {
+  if (response === "Update Settings") {
+    vscode.commands.executeCommand("workbench.action.openSettings");
+  }
 }
 
 /**
  * Execute a command of Yseop CLI using the folder from which VSCode has been opened as the KB directory.
- * @param yseopCliPath The absolute path to Yseop CLI executable.
+ * @param cliPath The absolute path to Yseop CLI executable.
```
mminot-yseop: Did the linter yell on this name for some reason?

alacambre-yseop: In fact `yseopCliPath` was used both as a global variable and an argument.
So yeah, it yelled because of the name shadowing.

alacambre-yseop:
The two-spaces indentation is hard to read in this context, but in VSCode it's OK thanks to the lines showing the indentation.

PR #64, opened 2019-02-13
feat: improved grammar; less (no?) ambiguities;

alacambre-yseop:
Should parse the majority of YML files completely, except text granules (content is skipped).
Assumes that devs use a semi-colon at the end of most of the instructions.
→ The extension now displays warnings for parsing problems.

PR #65, opened 2019-02-15
feat: add completion items for instances and functions

alacambre-yseop:


alacambre-yseop:
Waiting for #64 to be merged.

alacambre-yseop:
OK for review @mminot-yseop .

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -0,0 +1,247 @@
...
+      const result = parser.kaoFile();
+      const completionItems: CompletionItem[] = [];
+      const visitor = new YmlKaoFileVisitor(completionItems);
+      visitor.visit(result);
+      const expectedCompletionItems = [
+        {
+          data: "id_City_name",
+          detail: "String",
+          documentation: "not documented",
+          kind: 10,
```
mminot-yseop: If the kind is (I might be wrong) an auto-generated Antlr token number or something of this kind that might change, checking it might make your test needlessly brittle. Any opinion on this?

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -0,0 +1,247 @@
...
+      assert.equal(tokens[0].startIndex, 0);
+      assert.equal(tokens[0].stopIndex, 8);
+      done();
+    });
+
+    it("should parse a well written YML class and provide completion for fields", (done) => {
+      const inputStream = new ANTLRInputStream(`
+            interface City
+                field name
+                --> domains	String
```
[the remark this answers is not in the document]
mminot-yseop: Also true for very similar lines further down.

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -0,0 +1,247 @@
...
+      const lexer = new YmlToBdlLexer(inputStream);
+      const tokenStream = new CommonTokenStream(lexer);
+      const tokens = tokenStream.getTokens();
+      assert.equal(tokens.length, 1);
+      assert.equal(tokens[0].startIndex, 0);
+      assert.equal(tokens[0].stopIndex, 8);
+      done();
+    });
+
+    it("should parse a well written YML class and provide completion for fields", (done) => {
```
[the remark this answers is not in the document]
mminot-yseop: (multiple occurrences)

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -0,0 +1,247 @@
...
+        ;
+
+        Collection collectionWithLevel2
+        --> domainsLevel2 Text
+        ;
+
+        function functionWithoutArgsWithPar(Object arg1, Text arg2)
+          --> domains Text
+          --> return "it works"
+        ;
```
mminot-yseop: Dumb detail: you have

```
foo()
foo args {}
foo args { a b }
foo(a, b)
```

and I expected

```
foo()
foo args {}
foo(a, b)
foo args { a b }
```

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -0,0 +1,247 @@
...
+        },
+        {
+          data: "id_static_functionWithoutArgsWithPar",
+          detail: "Text",
+          documentation: "not documented",
+          kind: 3,
+          label: "functionWithoutArgsWithPar",
+        },
+      ];
+      assert.strictEqual(completionItems.length, 7);
```
mminot-yseop: OK if `deepEqual` provides a crappy message if ≠ in length. Unnecessary if it handles that in a nice way. I think. Dunno. Do as you wish.

server/src/visitors/YmlClassVisitor.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(completionItems: CompletionItem[]) {
+    super(completionItems);
+  }
+
+  public visitMethodDeclaration(node: MethodDeclarationContext): any {
+    createNewCompletionItem(
+      this.completionItems,
+      node.methodIntro().ymlId(),
+      node.field(),
+      CompletionItemKind.Method,
```
mminot-yseop: Is that where the numbers seen in the tests come from? Is there a way to get something roughly equivalent to an enum value's name in Java? ⇒ `Method` rather than an integer?

server/src/visitors/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+  let domainsLevel2 = "";
+  try {
+    for (const element of fieldOptions.filter((elem) => !!elem.commonField)) {
+      const option = element.commonField();
+      const optionName = option._optionName.text;
+      if (optionName === "domains") {
+        domains = option._optionValues[0].text;
+      } else if (optionName === "domainsLevel2") {
+        domainsLevel2 = ` − ${option._optionValues[0].text}`;
+      }
```
mminot-yseop: Could you add a final empty `else` clause containing a comment explaining which cases are contained in it? Like “YML fields unrelated to documentation.”?

mminot-yseop:
I'll test this in a while, I guess. Might be fun.

alacambre-yseop:
The next step is to map each suggestion to its declaring file URI.
This would allow us to avoid duplication in the suggestions list and to maybe add the `go to definition` feature.
Also, we can start to work on a way to access local variables or arguments within methods and functions.

mminot-yseop:
I have lots of

```
[Error - 13:54:53] [object Object]
[Error - 13:54:53] [object Object]
[Error - 13:54:53] [object Object]
[Error - 13:54:53] [object Object]
```

in the output, which are a bit scary and not really helpful xD But might be from develop; I'll check.

Hum… I got the class `==LibDocument:NlgEntity` in my completion items. I didn't expect those equal signs to sneak in like this. Not sure where they come from.

I succeeded in getting functions name to show up, but I'm not sure how that works for instances. I thought having

```
implementation GeneralData
    --> hasAccessorFunction true
;
```

would allow me to autocomplete “GeneralData” but I tried to no avail. :thinking:
Please provide some guidance xD

mminot-yseop:
I confirm I do not have this flood of errors on the output for `develop`. Consider grouping some of these errors (“× 3: foo” rather than “foo foo foo”, for example, I dunno), making them less frightful, or making them more informative if possible. Or several of these things together.

alacambre-yseop:
Thanks for testing.
> I have lots of
>
```
[Error - 13:54:53] [object Object]
[Error - 13:54:53] [object Object]
[Error - 13:54:53] [object Object]
[Error - 13:54:53] [object Object]
```

This is due to a NPE + instead of printing the error message we print the error itself which leads to this weird `[object Object]` thing.

> Hum… I got the class `==LibDocument:NlgEntity` in my completion items. I didn't expect those equal signs to sneak in like this. Not sure where they come from.

Yeah, this is a side effect of unperfect parsing. We need to improve the parser.

> --> hasAccessorFunction true

have no specific meaning for us. It is used only by YBS.
Here is a simple example of object instance:
`Symbol _NO_PASSIVE_FORM;`

There are other examples in the tests provided in this PR.

 > I confirm I do not have this flood of errors on the output for `develop`. Consider grouping some of these errors (“× 3: foo” rather than “foo foo foo”, for example, I dunno), making them less frightful, or making them more informative if possible. Or several of these things together.

I didn't find a way to do this with current VSCode API.

mminot-yseop:
> > Hum… I got the class `==LibDocument:NlgEntity` in my completion items. I didn't expect those equal signs to sneak in like this. Not sure where they come from.
> 
> Yeah, this is a side effect of unperfect parsing. We need to improve the parser.

As a temporary measure, we could trim completion items by removing leading and trailing characters that cannot be part of an YML identifier. :thinking: Wadayathink?

> > --> hasAccessorFunction true
> 
> have no specific meaning for us. It is used only by YBS.

Yeah, yeah, I was just copypasting the whole thing. I was mostly referring to the implementation thingy.

> Here is a simple example of object instance:
> `Symbol _NO_PASSIVE_FORM;`

Oh, I tested that earlier, kinda at random.

> > I confirm I do not have this flood of errors on the output for `develop`. Consider grouping some of these errors (“× 3: foo” rather than “foo foo foo”, for example, I dunno), making them less frightful, or making them more informative if possible. Or several of these things together.
> 
> I didn't find a way to do this with current VSCode API.

Omg, so what should we do? Can we prevent the NPE from happening or something? The current state is weird.

alacambre-yseop:
> Omg, so what should we do? Can we prevent the NPE from happening or something? The current state is weird.

I added a commit to fix the NPE.

> As a temporary measure, we could trim completion items by removing leading and trailing characters that cannot be part of an YML identifier.  Wadayathink?

I'd rather leave it like this and fix the parser ASAP.

mminot-yseop:
> I'd rather leave it like this and fix the parser ASAP.

Kay if you think it's not too hard.

Issue #66, opened 2019-02-19
[feature] Go to definition


alacambre-yseop:
Since the parser in greatly inproved, that would be great to have the `Go to definition` feature implemented.

PR #70, opened 2019-02-22
feat: go to definition and peek definition

alacambre-yseop:
See #66 
~Needs PR #69 to be merged first~


server/src/server.ts
```diff
@@ -135,6 +145,17 @@ function validateTextDocument(textDocument: TextDocument): void {
   }
 }
 
+connection.onDefinition((pos: TextDocumentPositionParams) => {
+  const doc: TextDocument = documents.get(pos.textDocument.uri);
+  const linePrefix = doc.getText().substr(0, doc.offsetAt(pos.position));
+  const lastToken = linePrefix.match(/\b([^:\.\s]+)\b$/g);
```
mminot-yseop: * You might have a “whole word” flag available to avoid explicitly using `\b` on both sides.
* The `\b$` at the end seems a bit redundant, don't you think? :thinking: Well, actually, it does not mean _exactly_ the same thing than just `$`:

image

Indeed, to have a “boundary”, you need the last character to be a “word character”.
_Shikashi_, do you _want_ this distinction to apply here? Do you _need_ a word character to appear at the end of your `[^:\.\s]+`? I'd consider doing one of these things:

* turning `[^:\.\s]+` into `[^:\.\s]*\w` and getting rid of the `\b` to state explictly that you need this thing to and with a word character. But I'm not even sure you need one; I'll let you tell me.
* just getting rid of the final `\b` in favor of a lone `$`, if you do not need a word character at the end.

### By the way

Why `\.` within the character list? Most metacharacters do not need to be escaped when they appear in this context. Generally, only square brackets cause a bit of trouble.

image


### By the way again

What is this capturing group (`( … )`) for?

server/src/server.ts
```diff
@@ -135,6 +145,17 @@ function validateTextDocument(textDocument: TextDocument): void {
   }
 }
 
+connection.onDefinition((pos: TextDocumentPositionParams) => {
+  const doc: TextDocument = documents.get(pos.textDocument.uri);
+  const linePrefix = doc.getText().substr(0, doc.offsetAt(pos.position));
+  const lastToken = linePrefix.match(/\b([^:\.\s]+)\b$/g);
+  if (lastToken === null || lastToken.length === 0) {
```
mminot-yseop: It's confusing to have an array called something like “lastFoo” since it sounds like there can be only one such thing. And indeed you seem to be expected at most one element. But that still sounds weird and unnatural.

server/src/visitors/YmlClassVisitor.ts
```diff
@@ -22,6 +27,14 @@ export class YmlClassVisitor extends YmlToBdlBaseVisitor {
       CompletionItemKind.Method,
       this.classId,
     );
+    this.definitions.addDefinition(
```
mminot-yseop: You seem to be always calling this `addDefinition` method along with a `createLocation` call. Why not change `addDefinition`'s parameter type and accept this `node` thingy and an URI? :thinking:

`addDefinition(node, this.uri);`

server/src/visitors/YmlClassVisitor.ts
```diff
@@ -22,6 +27,14 @@ export class YmlClassVisitor extends YmlToBdlBaseVisitor {
...
+      createLocation(
+        node.methodIntro().ymlId().text,
+        node.start,
+        node.stop,
+        this.uri,
+      ),
```
mminot-yseop: Surprised by this trailing comma since it does not seem to be a list or anything, and the function only seems to accept one location at a time anyway. :thinking:

server/grammar/Yml.g4
```diff
@@ -123,17 +124,12 @@ LINE_COMMENT: '//' ~[\n\r]* -> channel(HIDDEN);
...
 /*
  * Parser rules
  */
-
-expressionMarker: DOT | MULTIVALUED_EXPRESSION;
-
-ymlId: YMLID;
-
 // Base rule when parsing a file. Basicaly, this describes all valid YML files.
+// misses the `project.kao`-like files
```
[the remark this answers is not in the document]
alacambre-yseop: No, they are completely not handled yet and we have warnings because we're not able to parse them.

server/src/definitions/YmlDefinitionProvider.ts
```diff
@@ -0,0 +1,50 @@
...
+      return null;
+    }
+    const defs = this.definitions
+      .filter((defLoc) => defLoc.entityName.indexOf(token) >= 0)
+      .map((defLoc) => defLoc.location)
+      .reduce((prev: Location[], elem) => {
+        prev.push(elem);
+        return prev;
+      }, []);
+    if (!defs || defs.length === 0) {
```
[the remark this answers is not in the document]
alacambre-yseop: > Is `!defs` ⇔ `defs != null`?

Yes.

> And _can it_ be `null` at this point?

It is not possible, indeed.

server/src/server.ts
```diff
@@ -112,19 +118,23 @@ function validateTextDocument(textDocument: TextDocument): void {
...
   const result = parser.kaoFile();
+  defintionsProvider.removeDocumentDefinitions(textDocument.uri);
 
-  const visitor = new YmlKaoFileVisitor(completionItems);
+  const visitor = new YmlKaoFileVisitor(
+    completionItems,
+    textDocument.uri,
+    defintionsProvider,
+  );
   visitor.visit(result);
```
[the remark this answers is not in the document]
alacambre-yseop: The `visit` method is provided by ANTLR and is kind of the main method.

server/src/server.ts
alacambre-yseop: I will rename this one.

server/src/visitors/YmlClassVisitor.ts
alacambre-yseop: I will try with the node only. Good advice.

server/src/visitors/YmlClassVisitor.ts
alacambre-yseop: Autoformatting. Seems to be a good practice in TS :open_mouth:

mminot-yseop: Oh, I guess it's for when you refactor and add an argument to the method. Then it behaves kinda like when you add items to a list or array.

server/grammar/Yml.g4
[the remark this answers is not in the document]
mminot-yseop: Write what you want, but the “misses” sounds weird and is ambiguous and you should capitalize the sentence and add a period.

server/src/visitors/YmlClassVisitor.ts
alacambre-yseop: Unfortunately, after doing that I had cross-references problem that made execution fails.

server/src/definitions/DefinitionUtils.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Find in a text the last substring that could be a valid YML entity name's substring.
+ * @param text Some text.
+ * @returns the canditate entity's name or null if nothing could be found.
+ */
+export function getDocumentLastWord(text: string) {
+    if (!text) {
+      return null;
+    }
+    const matchArray = text.match(/\b([^:\.\s]+)\b$/g);
```
mminot-yseop: I have no idea why I didn't mention it earlier, but… shouldn't we extract the regex to avoid compiling it multiple times? Especially now that this is in a file that already contains some regexp constants. :slightly_smiling_face:

server/src/definitions/DefinitionUtils.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Find in a text the last substring that could be a valid YML entity name's substring.
+ * @param text Some text.
+ * @returns the canditate entity's name or null if nothing could be found.
+ */
+export function getDocumentLastWord(text: string) {
```
mminot-yseop: Confusing name since you don't really use it on whole “documents” and look for YML IDs rather than real words.

alacambre-yseop:
> * Weren't you including colons in the previous version?

Maybe I did, but I don't want to.

server/src/definitions/DefinitionUtils.ts
```diff
@@ -0,0 +1,24 @@
+/* Match valid YML entity names substrings:
```
mminot-yseop:
```suggestion
/**
 * Match valid YML entity names substrings:
```

?

PR #74, opened 2019-02-25
feat: create a completion provider

alacambre-yseop:


mminot-yseop:
This looks more like refactoring. Are there actual behavioural changes to expect?

alacambre-yseop:
This adds a way of resetting the completion list before parsing a file again.
There will be more things built on top of it, like a more contextualized completion (local variables and args in a function, for example).

PR #77, opened 2019-03-01
feat(completion): give access to local variables and functions arguments

alacambre-yseop:


client/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Release Notes
 
+## 1.4.0 (YYYY-MM-DD)
+
+- [BETA] Local variables and arguments are suggested when using autocompletion in functions.
+  - For loops local variables as well as local variables declared using the arrow operator (`--> local […]`) are not handled yet.
```
mminot-yseop: I had a hard time reading this, mostly because I thought this initial “for” was the “real” English “for”. Also not that English guys tend to use “…” without the square brackets.

```suggestion
  - Variables declared within “for” loops, as well as those declared using the arrow operator (`--> local …`), are not handled yet.
```

server/src/completion/ICompletionItemLocation.ts
```diff
@@ -12,4 +12,14 @@ export interface ICompletionItemLocation {
      * The completion information.
      */
     completion: CompletionItem;
+    /**
+     * The first position from which we can suggest this item.
+     * First character of a function, for example.
+     */
```
mminot-yseop: I had to check `getAvailableCompletionItems` to understand those comments. :sob:

```suggestion
     * Useful to avoid giving completion suggestions pertaining to symbols
     * that are out of the current scope.
     */
```

server/src/completion/ICompletionItemLocation.ts
```diff
@@ -12,4 +12,14 @@ export interface ICompletionItemLocation {
      * The completion information.
      */
     completion: CompletionItem;
+    /**
+     * The first position from which we can suggest this item.
+     * First character of a function, for example.
```
mminot-yseop:
```suggestion
     * First character of a function's body, for example.
```

server/src/completion/ICompletionItemLocation.ts
```diff
@@ -12,4 +12,14 @@ export interface ICompletionItemLocation {
...
+     */
+    scopeStartOffset?: number;
+    /**
+     * The last position to which we can suggest this item.
+     * Last character of a function, for example.
```
mminot-yseop: Kinda same issues with those comments. ^^

server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -26,7 +26,23 @@ export class YmlCompletionItemsProvider {
...
+          // We have the scope's positions set and are in the correct file.
+          if (offset < elem.scopeStartOffset || offset > elem.scopeEndOffset) {
+            // Outside of the scope.
+            return null;
+          } else {
+            return elem.completion;
+          }
+        } else {
+          // No information about the scope. The element is available everywhere.
+          return elem.completion;
```
mminot-yseop: * It surprises me that you have two nested if-else with the very same “else” content. You could probably merge stuff. If it results in weird huge conditions, use functions (well, or just one or whatever) to make sure people understand what's happening.
* As a general rule, if your “else” is way shorter, invert your if-else to increase readability.

server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -26,7 +26,23 @@ export class YmlCompletionItemsProvider {
...
+        }
+      })
+      // Remove null elements.
+      .filter((elem) => !!elem);
```
mminot-yseop: Silly question: is there an equivalent of Java's flatMap to avoid having to manually remove stuff afterwards? Like, instead of sprouting `null` values, you could return empty streams. Dunno.

mminot-yseop:
:eyes:

server/src/visitors/YmlFunctionVisitor.ts
```diff
@@ -1,11 +1,29 @@
...
+  VariableBlockContentContext,
+} from "../grammar";
 import { createLocation, createNewCompletionItem } from "./VisitorsUtils";
 import { YmlBaseVisitor } from "./YmlBaseVisitor";
 
 export class YmlFunctionVisitor extends YmlBaseVisitor {
+  /**
+   * First character of the Function object.
+   * Will be used as the start of the scope for its arguments and local variables.
+   */
```
mminot-yseop: Silly question again: Why are there similar pieces of information here and in the `ICompletionItemLocation` thingy from above?

server/src/visitors/YmlFunctionVisitor.ts
```diff
@@ -28,13 +48,70 @@ export class YmlFunctionVisitor extends YmlBaseVisitor {
...
-       * The function is a method instance.
-       * It has already been added as a completion item and a definition by YmlClassVisitor.
-       */
+        /*
+        * The function is a method instance.
+        * It has already been added as a completion item and a definition by YmlClassVisitor.
+        */
+    }
+    /*
+     * We need to keep track of the function position in the document
```
mminot-yseop: :yum: 

```suggestion
     * We need to keep track of the function position in the document.
```

server/src/visitors/YmlFunctionVisitor.ts
```diff
@@ -28,13 +48,70 @@ export class YmlFunctionVisitor extends YmlBaseVisitor {
...
+     * This allows us to retreive its arguments and local variables.
```
mminot-yseop:
```
$ dico-en retreive
retrieve
```

```suggestion
     * This allows us to retrieve its arguments and local variables.
```

server/src/visitors/YmlFunctionVisitor.ts
```diff
@@ -28,13 +48,70 @@ export class YmlFunctionVisitor extends YmlBaseVisitor {
...
+    createNewCompletionItem(
+      this.uri,
+      this.completionProvider,
+      node._argName.text,
+      [],
+      CompletionItemKind.Variable,
+      this.functionName,
+      node._argType.text,
+      this.scopeStartOffset,
+      this.scopeEndOffset,
```
mminot-yseop: The number of arguments is starting to be a little frightening :fearful:
Any idea to lower that? Is there some kind of _range_ class for the start / end thingies, for example?

Note that using a builder probably would not be a(n elegant) solution if 99 % of those things are mandatory.

server/src/visitors/YmlFunctionVisitor.ts
```diff
@@ -49,10 +126,7 @@ export class YmlFunctionVisitor extends YmlBaseVisitor {
     const functionNameSubParts = fullName.split("::");
     if (functionNameSubParts.length > 1) {
       methodName = functionNameSubParts.pop();
-      className = fullName.replace(
-        new RegExp(`::${methodName}$`, "u"),
-        "",
-      );
+      className = fullName.replace(new RegExp(`::${methodName}$`, "u"), "");
```
mminot-yseop: * Old code, but still, aren't we compiling a regexp several times for nought, here?
* On my Prism.js PR, I was told (in js) that `RegExp(…)` ⇔ `new RegExp(…)` and that the shorter form was preferred. See if it works the same way in ts.

server/src/completion/YmlCompletionItemsProvider.ts
alacambre-yseop: This is experimental and doesn't seem to work :(

server/src/visitors/YmlFunctionVisitor.ts
alacambre-yseop: I prefer the version with because of `Java` (bad?) habits. It makes clearer to me that we are creating a new object.
The regexp is dynamic. We then compile it only once for each function/method.

server/src/visitors/YmlFunctionVisitor.ts
alacambre-yseop: This is one of the future development steps: create a real model.
Now we are just dynamically building the completion items and definition items, but it should be cleaner to have real objects, like `YmlFunction` for example, that would contain there children objects.

server/src/visitors/YmlFunctionVisitor.ts
alacambre-yseop: This duplicated elements will be removed when a real model be created.

server/src/visitors/YmlFunctionVisitor.ts
mminot-yseop: Hm. But note that if this were on the Suite, with Sonar and stuff, you wouldn't be able to merge because you crossed the “seven arguments” boundary xD

Issue #78, opened 2019-03-05
“local” keyword highlighting


mminot-yseop:
When writing functions, the `local` keyword used for the local variable declaration block is not highlighted, whereas the `static` one is.

![image](https://user-images.githubusercontent.com/36635838/53794934-3106fe00-3f31-11e9-9f74-111e8065674f.png)

### Expected
Similar style for `local` and `static` in this context.

### What happens
`local` has the default colour, while `static` is OK.

### Sample code

```
function printXMLSchema(Class MyClass)
local {
	Integer foo
}
static {
	StringBuffer sb
		--> maxSize 100000
	;
}
--> local XmlStream xs
--> domains String
--> action {
	xs = new(XmlStream);
}
;
```

alacambre-yseop:
@mminot-yseop I close this issue since version 1.5.0 fixes it.

PR #83, opened 2019-03-11
fix: syntax highlighting

alacambre-yseop:
See https://github.com/yseop/vscode-yseopml/issues/78

### DIFF
--- a/client/syntaxes/yml.tmLanguage.json
+++ b/client/syntaxes/yml.tmLanguage.json
@@ -24,7 +24,7 @@
         "keywords": {
             "patterns": [{
                     "name": "keyword.control.yml",
-                    "match": "\\b(if|while|for|return|foreach|switch|switchExpr|switchExprExclusive|belongsTo|default|case|interface|class|implementation|private|extends|override|complete|field|method|function|static|classProperties|ifExpr|then|else)\\b",
+                    "match": "\\b(args|belongsTo|break|case|class|classProperties|complete|default|else|extends|field|for|foreach|function|Function|if|ifExpr|implementation|interface|local|method|override|private|return|static|switch|switchExpr|switchExprExclusive|then|while)\\b",
                     "comment": "All YML keywords that should be highlighted."
                 },
                 { "include": "#comments" },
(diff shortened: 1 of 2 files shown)


client/syntaxes/yml.tmLanguage.json
```diff
@@ -24,7 +24,7 @@
         "keywords": {
             "patterns": [{
                     "name": "keyword.control.yml",
-                    "match": "\\b(if|while|for|return|foreach|switch|switchExpr|switchExprExclusive|belongsTo|default|case|interface|class|implementation|private|extends|override|complete|field|method|function|static|classProperties|ifExpr|then|else)\\b",
+                    "match": "\\b(if|while|for|return|foreach|switch|switchExpr|args|switchExprExclusive|belongsTo|default|case|interface|class|implementation|private|extends|override|complete|field|method|function|Function|static|local|classProperties|ifExpr|then|else|break)\\b",
```
mminot-yseop: The order is getting more and more confusing. :thinking:

```
$ tr '|' '\n' <<< 'if|while|for|return|foreach|switch|switchExpr|args|switchExprExclusive|belongsTo|delt|case|interface|class|implementation|private|extends|override|complete|field|method|function|Function|static|local|classProperties|ifExpr|then|else|break' | sort --ignore-case | tr '\n' '|' | sed 's,|$,\n,'
args|belongsTo|break|case|class|classProperties|complete|default|else|extends|field|for|foreach|function|Function|if|ifExpr|implementation|interface|local|method|override|private|return|static|switch|switchExpr|switchExprExclusive|then|while
```

```suggestion
                    "match": "\\b(args|belongsTo|break|case|class|classProperties|complete|default|else|extends|field|for|foreach|function|Function|if|ifExpr|implementation|interface|local|method|override|private|return|static|switch|switchExpr|switchExprExclusive|then|while)\\b",
```

My work on Prism.js made me realize that once lists grow to a certain size, the alphabetic order is generally more helpful than most attempts at semantical organization.

mminot-yseop:
Test OK. Waiting for opinions regarding my previous comment.

mminot-yseop:
Still seems OK, but you upset the linter.

PR #84, opened 2019-03-14
feat: add ids or ids subparts as text completion items

alacambre-yseop:


client/CHANGELOG.md
```diff
@@ -2,77 +2,78 @@
 
 ## 1.X.X (YYYY-MM-DD)
 
-- Fix syntax highlighting for keywords `local`, `args`, `Function` and `break`.
-- Improved completion from Yseop Engine's model. More elements are available such as:
-  - functions' names;
-  - text tags;
-  - methods;
-  - attributes.
+-   Fix syntax highlighting for keywords `local`, `args`, `Function` and `break`.
```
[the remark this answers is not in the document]
mminot-yseop: I kinda understand, but since people also have to work with old code and stuff, it's probably better to highlight both.

server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -8,11 +8,28 @@ export class YmlCompletionItemsProvider {
...
      */
     public addCompletionItem(complLoc: AbstractYmlObject): void {
         this.completions.push(complLoc);
+        const subElements = complLoc.label.split(/::?/);
+        /* Single element. This has already been added. */
+        if (subElements.length === 1) {
+            return;
+        }
+        /* Keep all subparts of the object's label, even its short name. */
+        for (const word of subElements) {
```
mminot-yseop: I'd check for non-blankness, since a typo could lead to a `:::::::`, which would make `split` yield empty elements.
That, or turn the regex into `:+`.

server/src/server.ts
```diff
@@ -50,7 +50,7 @@ connection.onInitialize(
             capabilities: {
                 // Tell the client that the server support code complete
                 completionProvider: {
-                    resolveProvider: true,
+                    triggerCharacters: ['.', ':'],
```
mminot-yseop: I guess it would have a poor “simplicity to usefulness” ratio, but I don't like the fact that if you

1) type two colons and
2) use backspace to erase one

the suggestions disappear and you have to hit Ctrl-Space yourself xd

mminot-yseop:
image
:thinking: :cry:

server/src/server.ts
alacambre-yseop: What trigger characters should I set?

server/src/completion/YmlCompletionItemsProvider.ts
alacambre-yseop: Done.

client/CHANGELOG.md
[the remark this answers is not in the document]
ahelman-yseop: or force them change `Function` to `function` to have a nice color :D 
(JK, I kinda understand)

server/src/server.ts
mminot-yseop: I dunno, I was just pointing at the fact that this system wasn't perfect for our case, but I doubt it would be easy to fix that, nor that it would be worth the trouble.

server/src/completion/YmlCompletionItemsProvider.ts
mminot-yseop: I realized a bit late that the `:+` technique can still produce empty elements if the colons stand at the end of the string (i.e., when we haven't typed what follows yet). However:

* maybe you only call this method with things that match a pattern avoiding this?
* I'm not even sure it's _that_ awful to add empty elements.

Still, to avoid unnecessary questions and make things look more… natural, I'd probably use the `:+` _and_ a `if not blank` together, if I were you.

alacambre-yseop: - Yes, indeed, I'm calling this method only when we encounter an YML object that has a correct identifier.
Identifiers cannot finish with a colon, it should be OK.
- I think this is awful :smile: 

I will add a `if not blank` condition on the `word` variable.

server/src/server.ts
alacambre-yseop: I will keep the dot alone, then.

mminot-yseop: Wut? No! The colon is helpful, dude. What are you talking about? xD

server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -8,11 +8,35 @@ export class YmlCompletionItemsProvider {
...
+            /* Can be empty if colon at the end of the label. */
+            if (word.length === 0) {
+                continue;
+            }
+            const keyword = new TextTokenKeyword(word);
+            if (this.getItem(keyword.data)) {
+                continue;
```
mminot-yseop:
```suggestion
                /* Already available. */
                continue;
```

mminot-yseop:
Test OK for the latest version. Just consider adding the suggested comment and I'm done.

alacambre-yseop:
@mminot-yseop I added the comment as suggested :smile:

PR #85, opened 2019-03-14
feat(grammar): improve parsing; add forall; add try/catch; complete

alacambre-yseop:


mminot-yseop:
LOL I don't understand where the manual changes are (except for the warning → error thing). Everything looks like it was automatically generated.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+
+granule:
+    OPEN_GRANULE (~(OPEN_GRANULE | CLOSE_GRANULE)+ | granule)*? CLOSE_GRANULE EOF?
+;
+
+objectComplete:
+    COMPLETE completedElemId=ymlId memberOption=field* SEMICOLON
+;
+
+classComplete:
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+;
+constList:
+    OPEN_BRACE elements+=value? (COMMA elements+=value)* CLOSE_BRACE
+;
+
+granule:
+    OPEN_GRANULE (~(OPEN_GRANULE | CLOSE_GRANULE)+ | granule)*? CLOSE_GRANULE EOF?
+;
+
+objectComplete:
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+    | instruction_return
+    | instruction_ifElse
+    | instruction_try_catch SEMICOLON?
+    | instruction_switchCase_asIf
+    | instruction_break
+    | instruction_switchCase_withValue
+    | instruction_ifExprBlock
+    | instruction_while
+;
+instruction_do: 'do' actionBlock;
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+instruction_try_catch:
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+instruction_while: WHILE OPEN_PAR order0Condition CLOSE_PAR actionBlock;
+instruction_return: RETURN value SEMICOLON?;
+instruction_chainedCall: chainedCall;
+instruction:
+    instruction_chainedCall SEMICOLON?
+    | instruction_multivaluedAssignment SEMICOLON?
+    | instruction_assignment SEMICOLON?
+    | instruction_for
+    | instruction_forEach
+    | instruction_forall
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+    | instruction_return
+    | instruction_ifElse
+    | instruction_try_catch SEMICOLON?
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+    CASE OPEN_PAR valueOrCondition CLOSE_PAR COLON (instruction | actionBlock)
+;
+instructionDefault: DEFAULT COLON (instruction | actionBlock);
+instruction_break: BREAK SEMICOLON?;
+
+instruction_ifElse: instruction_if (ELSE (actionBlock | instruction))?;
+instruction_if:
+    IF OPEN_PAR order0Condition CLOSE_PAR (actionBlock | instruction)
+;
+instruction_forall:
```
alacambre-yseop: Manual edit here.

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+    FORALL OPEN_PAR (ymlId | instanciationVariable) IN (value | FUNCTION) CLOSE_PAR
```
mminot-yseop: I was expecting `functionCall` rather than `FUNCTION` :thinking: So you can write something like…

```
forall (foo in function)
```

?

:thinking:

server/grammar/Yml.g4
```diff
@@ -1,458 +1,482 @@
...
+    | instruction_try_catch SEMICOLON?
+    | instruction_switchCase_asIf
+    | instruction_break
+    | instruction_switchCase_withValue
+    | instruction_ifExprBlock
+    | instruction_while
+;
+instruction_do: 'do' actionBlock;
+instruction_try_catch:
+    'try' OPEN_PAR instruction_do 'catch' OPEN_PAR (ymlId (COMMA ymlId)*?) CLOSE_PAR actionBlock CLOSE_PAR
```
mminot-yseop: No allcaps-token-constant-thingies for “do”, “try” and “catch”? Because just one occurrence? It's surprising since in this file you can scroll for hours without seeing any hard-coded string and then PAM, “do try catch”.

server/grammar/Yml.g4
alacambre-yseop: No, in this case, we do accept the `FUNCTION` token too. This is clearly intended.

server/grammar/Yml.g4
alacambre-yseop: Woops. It seems I was lazy. Sorry. I'll fix that.

server/grammar/Yml.g4
mminot-yseop: Then I'd like to see an YML code example, because this is puzzling xD

alacambre-yseop: @mminot-yseop I explained what it does.

mminot-yseop: I'm starting to understand, but why are functions the _only_ type that needs to be explicitly mentioned in this rule? :thinking: What about `Class`, for example, or stuff like this?

alacambre-yseop: We don't have a token explicitly defined for `Class`, while we have it for `Function`. Therefore the type of the token receiving the text `Function` will be of type `FUNCTION` and not of type `YMLID` as expected in the rule `value`.

mminot-yseop: So `Class` will count as a “value” and match that part of the rule, is that it?

mminot-yseop: And the use of `FUNCTION` just adds information to the syntactic tree that get built?

alacambre-yseop: > So Class will count as a “value” and match that part of the rule, is that it?

Yes, exactly.

> And the use of FUNCTION just adds information to the syntactic tree that get built?

If the parser doesn't have this information, the parsing will fail.

PR #86, opened 2019-03-15
feat: hovering known elements displays their documentation 

alacambre-yseop:
documentation

### DIFF
--- a/server/src/completion/YmlCompletionItemsProvider.ts
+++ b/server/src/completion/YmlCompletionItemsProvider.ts
@@ -67,7 +67,7 @@ export class YmlCompletionItemsProvider {
      * Find the *first* completion item that has `entityName` as name.
      *
      * @param entityName The name of the entity searched for.
-     * @returns The first completion item that has the same label as `entityName`, `null` otherwise.
+     * @returns The first completion item that has `entityName` as its label, `null` otherwise.
      */
     public getItemByLabel(entityName: string): AbstractYmlObject {
         return this.completions.find((elem) => elem.label === entityName);
--- a/server/src/completion/YmlCompletionItemsProvider.ts
+++ b/server/src/completion/YmlCompletionItemsProvider.ts
@@ -45,10 +45,6 @@ export class YmlCompletionItemsProvider {
      * @returns The first completion item that has the same label as `entityName`, `null` otherwise.
      */
     public getItemByLabel(entityName: string): AbstractYmlObject {
-        const sameNameElements = this.completions.filter((elem) => elem.label === entityName);
-        if (!sameNameElements || sameNameElements.length === 0) {
-            return null;
-        }
-        return sameNameElements.shift();
+        return this.completions.find((elem) => elem.label === entityName);
     }
 }
--- a/server/src/completion/YmlCompletionItemsProvider.ts
+++ b/server/src/completion/YmlCompletionItemsProvider.ts
@@ -37,4 +37,18 @@ export class YmlCompletionItemsProvider {
             })
             .filter((elem) => !!elem);
     }
+
+    /**
+     * Find the *first* completion item that has `entityName` as name.
+     *
+     * @param entityName The name of the entity searched for.
+     * @returns The first completion item that has the same label as `entityName`, `null` otherwise.
+     */
+    public getItemByLabel(entityName: string): AbstractYmlObject {
+        const sameNameElements = this.completions.filter((elem) => elem.label === entityName);
+        if (!sameNameElements || sameNameElements.length === 0) {
+            return null;
+        }
+        return sameNameElements.shift();
+    }
 }
--- a/server/src/definitions/DefinitionUtils.ts
+++ b/server/src/definitions/DefinitionUtils.ts
@@ -4,8 +4,12 @@
  * The whole name will be captured within a group.
  * see https://stackoverflow.com/a/51166092/3577898
  */
-const YML_ENTITY_NAME_START_REGEXP = /^((\p{Letter}|[_0-9])+(::?(\p{Letter}|[_0-9])+)*)/u;
-const YML_ENTITY_NAME_END_REGEXP = /((\p{Letter}|[_0-9])+(::?(\p{Letter}|[_0-9])+)*)$/u;
+
+const ALPHANUM_OR_UNDERSCORE = '(\\p{Letter}|[_0-9])';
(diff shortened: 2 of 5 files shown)


server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -37,4 +37,18 @@ export class YmlCompletionItemsProvider {
...
+
+    /**
+     * Find the *first* completion item that has `entityName` as name.
+     *
+     * @param entityName The name of the entity searched for.
+     * @returns The first completion item that has the same label as `entityName`, `null` otherwise.
+     */
+    public getItemByLabel(entityName: string): AbstractYmlObject {
+        const sameNameElements = this.completions.filter((elem) => elem.label === entityName);
+        if (!sameNameElements || sameNameElements.length === 0) {
```
mminot-yseop: ? I'd find it surprising if `filter(…)` could return `null` in some cases and an empty array / collection / whatever in other cases. :thinking:
:hourglass_flowing_sand: 

> A new array with the elements that pass the test. If no elements pass the test, an empty array will be returned.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -37,4 +37,18 @@ export class YmlCompletionItemsProvider {
...
+            return null;
+        }
+        return sameNameElements.shift();
```
mminot-yseop: Hum… Isn't there some way to avoid uselessly applying the filter on the whole array, since we end up only returning one element? I would even prefer a basic loop-based approach:

```
for(item in arr) {
    if(test) {
        return item;
    }
}

return null;
```

But maybe there are things in arrays like “findFirst” or whatever.

server/src/definitions/DefinitionUtils.ts
```diff
@@ -4,8 +4,8 @@
  * The whole name will be captured within a group.
  * see https://stackoverflow.com/a/51166092/3577898
  */
-const YML_ENTITY_NAME_START_REGEXP = /^((\p{Letter}|[_0-9])+)/u;
-const YML_ENTITY_NAME_END_REGEXP = /((\p{Letter}|[_0-9])+)$/u;
+const YML_ENTITY_NAME_START_REGEXP = /^((\p{Letter}|[_0-9])+(::?(\p{Letter}|[_0-9])+)*)/u;
+const YML_ENTITY_NAME_END_REGEXP = /((\p{Letter}|[_0-9])+(::?(\p{Letter}|[_0-9])+)*)$/u;
```
mminot-yseop: LOL you could probably factorize this stuff by using a string and the `RegExp` constructor.

https://github.com/PrismJS/prism/pull/1443/files#diff-43c117e56c7e839ab9e83577778c3236R9

It might involve adding a few backslashes, but nothing prohibitive, I think.

server/src/server.ts
```diff
@@ -60,6 +61,29 @@ connection.onInitialize(
     },
 );
 
+connection.onHover((_params) => {
+    const doc: TextDocument = documents.get(_params.textDocument.uri);
+    const entityName = getTokenAtPosInDoc(doc.getText(), doc.offsetAt(_params.position));
+    if (!entityName) {
+        return null;
+    }
+    console.warn(`search for ${entityName}`);
```
mminot-yseop: I'm confused by the fact that this is a warning. :thinking: Isn't this something perfectly normal?

server/src/server.ts
```diff
@@ -60,6 +61,29 @@ connection.onInitialize(
...
+
+    const entity = completionProvider.getItemByLabel(entityName);
+    if (!entity) {
+        return null;
+    }
+    if (!entity.documentation) {
```
mminot-yseop: LOL please use `||` to merge those two ifs.

mminot-yseop: and then maybe an explicit `else` to host the `return { stuff }`.

alacambre-yseop: I merged them, but I won't add an explicit `else`. Both indentation and end of block brace are enough I think.

server/src/server.ts
alacambre-yseop: Oops. I just didn't remove it back then.

server/src/completion/YmlCompletionItemsProvider.ts
alacambre-yseop: Done. The method `find` did the trick.

alacambre-yseop:
@ahelman-yseop I can explain whenever you want.

server/src/definitions/DefinitionUtils.ts
```diff
@@ -4,8 +4,12 @@
...
  * see https://stackoverflow.com/a/51166092/3577898
  */
-const YML_ENTITY_NAME_START_REGEXP = /^((\p{Letter}|[_0-9])+)/u;
-const YML_ENTITY_NAME_END_REGEXP = /((\p{Letter}|[_0-9])+)$/u;
+
+const ALPHANUM_OR_UNDERSCORE = '(\\p{Letter}|[_0-9])';
+const YML_ENTITY_NAME_REGEXP = `(${ALPHANUM_OR_UNDERSCORE}+(::?${ALPHANUM_OR_UNDERSCORE}+)*)`;
+
+const YML_ENTITY_NAME_START_REGEXP = RegExp(`^${YML_ENTITY_NAME_REGEXP}`, 'u');
+const YML_ENTITY_NAME_END_REGEXP = RegExp(`${YML_ENTITY_NAME_REGEXP}$`, 'u');
```
mminot-yseop: Oh, it's nice with those backticks. Dunno if they're available in pure JS too.

mminot-yseop:
Not your fault at all, but this doesn't get along with spell checkers very well xD
vsc-spell
I should try another spell-checking VSC extension some day. This one has quite a few annoying aspects.

server/src/completion/YmlCompletionItemsProvider.ts
```diff
@@ -62,4 +62,14 @@ export class YmlCompletionItemsProvider {
             })
             .filter((elem) => !!elem);
     }
+
+    /**
+     * Find the *first* completion item that has `entityName` as name.
+     *
+     * @param entityName The name of the entity searched for.
+     * @returns The first completion item that has the same label as `entityName`, `null` otherwise.
```
mminot-yseop: Not quite what's actually happening.

```suggestion
     * @returns The first completion item that has `entityName` as its label, `null` otherwise.
```

PR #89, opened 2019-03-20
chore(release): version 1.5.0

alacambre-yseop:


mminot-yseop:
:thinking: I was expecting a smaller diff. Is the release script messing with the JSON format or whatever?

alacambre-yseop:
That's possible, yeah. You're right, we should just have the version numbers changed as well as the date, nothing else.

mminot-yseop:
Note that I was assuming you were performing a few checks on the JSON, hence the lack of progress xD
What should we do?
Hum… Seems like the large diffs are due to indentation width changes. Should be harmless.
Still, we should try to find, in the future, what causes this xD

mminot-yseop:
https://github.com/yseop/vscode-yseopml/pull/90

alacambre-yseop:
Hum… We maybe have a problem here. GitHub asks us to update the branch, but this is clearly something we don't wanna do.
:thinking: 
We shouldn't merge #90 that early.

mminot-yseop:
Well, we can't really fix this now since this is on master, though. Should we bite the bullet for this particular release and endure the ugly indentation-changing diffs one last time?

mminot-yseop:
Hum. What should we do about the “update” thingy?

alacambre-yseop:
> Hum. What should we do about the “update” thingy?

I don't know :scream: .

mminot-yseop:
We could create a branch called… anything else than `master` (tracking `develop`), pull `master` to it, and merge that via a PR into develop.

mminot-yseop:
When you look at things calmly, it's nothing _that_ critical I think. We just need the changelog and the package.json numbers to land into `develop` somehow.

alacambre-yseop:
I don't feel very confident about this. You can try if you want.
The impacts, if any, should be limited as you said.
And it will be done using a PR, so…

alacambre-yseop:
Or we can ask OPS to disable the out-of-date check.

mminot-yseop:
> Or we can ask OPS to disable the out-of-date check.

I guess it would work, yeah, as long as we do not have conflicting files.

alacambre-yseop:
@mminot-yseop The out-of-date check has been completely removed by OPS.

mminot-yseop:
I hope there won't be any weird side-effects on the next release.

PR #91, opened 2019-03-21
feat: better message when engine model URI isn't a normal file

alacambre-yseop:


server/src/engineModel/EngineModel.ts
```diff
@@ -30,10 +30,12 @@ export class EngineModel {
      * Parse the file that is at `uri`.
      */
     public loadPredefinedObjects(): void {
-        if (fs.existsSync(this.uri)) {
-            this.parsePredefinedObjects(this.uri);
-        } else {
+        if (!fs.existsSync(this.uri)) {
             connection.console.warn(`File “${this.uri}” doesn't exist.`);
+        } else if (!fs.lstatSync(this.uri).isFile()) {
+            connection.console.warn(`It seems that ${this.uri} is not a file. Therefore it cannot be imported.`);
```
mminot-yseop: LOL this is actually _even more_ confusing, since technically in many systems like Linux people often say “everything is a file”. LOL there's even a Wikipedia page for that phrase. https://en.wikipedia.org/wiki/Everything_is_a_file
So…

```suggestion
            connection.console.warn(`It seems that ${this.uri} is not a normal file. Therefore, it cannot be imported.`);
```

:heavy_plus_sign: Maybe add to the message an indication like “Note that the expected path should point towards a `predefinedObjects.xml` file, not to an Yseop Engine directory.”, but then people will think “well, why don't you fetch the `predefinedObjects.xml` file automatically when I give an Engine directory path, then?” so maybe we should consider implementing that. I dunno.

alacambre-yseop: The feedback I received was more like “Why don't you take automatically the predefinedObjects.xml files depending on the version used by my KB?” xD

mminot-yseop: It would be possible by using the same environment variables as CLI, maybe, right? :thinking: or additional parameters. Or simply replacing the “predefined objects” parameter of the extension with a “engines directory” parameter.

mminot-yseop:
Well, still a bit vague but let's call it a day, I guess.

Issue #94, opened 2019-03-22
Parsing of implementation area


davanessoff-yseop:
I just give you an example and let you check the problems detected by the extension at the moment:
```
implementation SentenceToGenerate

	forGroups
	--> defaultValue theClinicalStudy.groupsOfSubjects
	
	override {
		write function
	}
;
```
I will give you other examples when I stumble upon some.

alacambre-yseop:
@davanessoff-yseop Fix should be available with #96.

PR #96, opened 2019-03-22
feat: attr impl in class impl part; class complete adding meth

alacambre-yseop:
see https://github.com/yseop/vscode-yseopml/issues/94


server/test/Parsing.test.ts
```diff
@@ -0,0 +1,52 @@
+import { CharStreams, CommonTokenStream } from 'antlr4ts';
+import * as assert from 'assert';
+
+import { YmlLexer, YmlParser } from '../src/grammar';
+
+describe('Parsing Tests', () => {
+    describe('classImplementation rule', () => {
+        it('should parse correctly implementation with attributes and override', (done) => {
+            const inputStream = CharStreams.fromString(`implementation SentenceToGenerate
```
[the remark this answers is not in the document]
alacambre-yseop: For this one, OK, but having the first line containing just an end-of-line character can cause problems in some other cases if we are sensitive to what's on the first line (which is not with our ANTLR grammar).
These strings have real meaning and I didn't choose to use an interpreted string just to have a beautiful file.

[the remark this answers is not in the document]
alacambre-yseop: @mminot-yseop Done

mminot-yseop:
> fix: use end-of-line to have more beautiful input strings

Lol, more like “chore”, “style” or “refactor” than “fix”, but OK xD

server/test/YmlKaoFileVisitor.test.ts
```diff
@@ -249,7 +253,7 @@ describe('Extension Server Tests', () => {
         });
         // tslint:disable-next-line: max-line-length
         it('should parse a well-written YML file containing instances and functions and provide completion for them', (done) => {
-            const inputStream = new ANTLRInputStream(`
+            const inputStream = CharStreams.fromString(`
```
mminot-yseop: By the way, what triggered this change?

server/test/Parsing.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    method manageUnitLabel(DataStructure localDataStructure, Text unitText {__nullable}) function
+    --> domains Void
+
+    method manageUnit(DataStructure localDataStructure) function
+    --> domains Void
+
+    method getStudyData(PatientGroupPartition group) function
+    --> domains StudyData
+;
```
mminot-yseop: Huuuum… How can we be sure this is not just parsed without error thanks to a stray greedy `.*` gobbling the whole thing up? :thinking: I guess adding tests like “it should not parse blahblah” with slightly altered (and incorrect) pieces of code would be kinda lame and cumbersome… And it would not prove much either.

server/test/YmlKaoFileVisitor.test.ts
alacambre-yseop: The upgrade of the ANTLR dependency makes `ANTLRInputStream` deprecated.

server/test/Parsing.test.ts
alacambre-yseop: > Huuuum… How can we be sure this is not just parsed without error thanks to a stray greedy .* gobbling the whole thing up?

I think we can't, but adding more tests on some other kinds of piece of codes can help.
Keep in mind that in some context the first solution we may implement can be a greedy globing thing.

mminot-yseop: We can't retrieve a rad tree or something? So disappointing…

Btw, “globbing” ≠ “gobbling”, hehe.

PR #107, opened 2019-07-02
feat(grammar): grammar improvments

alacambre-yseop:
- Added rules to the grammar;
- Improved some current rules;
- Fixed one NPE; 
- Added new parsing unit tests.

More to come in a next PR


server/grammar/Yml.g4
```diff
@@ -119,7 +119,7 @@ INTEGER: NUMBER+;
 // Colon are OK inside a YMLID
 // Colons aren't interpreted by YE in a particular way.
 YMLID: ID (COLON? COLON ID)*;
-ID: LETTER ALPHANUM*;
+ID: ALPHANUM* LETTER ALPHANUM*;
```
mminot-yseop: Wut? IDs starting with numerals? Intentional? YML?

server/grammar/Yml.g4
```diff
@@ -152,7 +152,7 @@ ymlEntity:
  */
 expressionMarker: DOT DOT | DOT | MULTIVALUED_EXPRESSION;
 
-ymlId: YMLID | ARGS | LOCAL | RETURN;
+ymlId: YMLID | ARGS | LOCAL | RETURN | FUNCTION_AS_TYPE | TEXT_FUNCTION;
```
mminot-yseop: Having `YMLID` and `ymlId` is confusing, btw. I’m having a hard time understanding what we’re dealing with, here.

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -29,6 +29,43 @@ describe('Extension Server Tests', () => {
             assert.equal(tokens.length, 1);
             assert.equal(tokens[0].startIndex, 0);
             assert.equal(tokens[0].stopIndex, 8);
+            assert.equal(tokens[0].type, YmlLexer.STRING);
+            done();
+        });
+
+        it('should be able to parse identifier with Japanese characters or accentuated characters', (done) => {
```
mminot-yseop:
```suggestion
        it('should be able to parse identifier with Japanese characters and diacritics', (done) => {
```

### OR

```suggestion
        it('should be able to parse identifiers with Japanese characters or diacritics', (done) => {
```

server/src/visitors/YmlFunctionVisitor.ts
```diff
@@ -38,7 +38,7 @@ export class YmlFunctionVisitor extends YmlBaseVisitor {
 
         if (!this.isMethodInstanciation(this.functionName)) {
             const func = new YmlFunction(this.functionName, this.uri);
-            func.enrichWith(node.field(), null);
+            func.enrichWith(node.field(), connection);
```
mminot-yseop: Change in the middle of nowhere? xD

server/grammar/Yml.g4
alacambre-yseop: Wep. It's completely valid. I added it on purpose.

server/grammar/Yml.g4
alacambre-yseop: I'm dealing with the tokens `Function` and `TextFunction` that can be used as an introduction keyword for a function, or as the type of a function's argument.

server/src/visitors/YmlFunctionVisitor.ts
alacambre-yseop: In some very rare cases, there was sometimes exceptions thrown because of this null.

alacambre-yseop:
> Gosh, complicated stuff.

Note that all files in `server/src/grammar/` are automatically generated.

server/grammar/Yml.g4
mminot-yseop: You misunderstood my comment. “What we’re dealing with, here” ⇔ “what this entity (ymlId) is”, not “what your change implies”. An ID that can be an argument list or a return or…? Wutwutwut

server/src/visitors/YmlFunctionVisitor.ts
mminot-yseop: :woman_shrugging: I hope this `connection` thingy makes sense.

mminot-yseop:
> > Gosh, complicated stuff.
> 
> Note that all files in `server/src/grammar/` are automatically generated.

Sure, lol, but these grammar edits, for someone that suck at YML…

server/grammar/Yml.g4
alacambre-yseop: The problem is that in YML most of the alpha-numeric expressions can be used as a valid identifier but at the same time YML has no reserved words. For example, I can have a local variable named `return`, `interface` or `function`, or even `Function`.
The only thing that makes sense for the YML compiler is the token order.
A specific token order has a specific meaning, but these tokens can still be used freely anywhere else.
That's the purpose of this ugly `ymlId` rule. It's more like a patch here. There is maybe a better solution, but I don't know it yet.

server/src/visitors/YmlFunctionVisitor.ts
alacambre-yseop: It is used to send an error log entry to the client when an exception occurs.

server/grammar/Yml.g4
mminot-yseop: But `ymlId` and `YMLID` seem very different from each other, hence the fact that I’m wondering if one of them could get a better name.

alacambre-yseop: YMLID (full caps) is a **lexer** rule. It explains how an id token is written.
ymlId is a **parser** rule. It accepts any YMLID and some other tokens that _could be_ YMLID but are more specialized tokens because I need them for some other **parser** rules.

mminot-yseop: Yeah, yeah, but the simple fact that there’s an homonymy between a lexer rule and a parser rule bugs me xD

mminot-yseop:
There’s not much I can say, here. Testing by writing random stuff would just replicate the unit tests you wrote. :/

server/grammar/Yml.g4
alacambre-yseop: Oh. I can either rename things right in this PR or do it in another, as you prefere.
The least impactful whould be to rename `YMLID` into `GENERIC_YMLID`.

PR #108, opened 2019-07-05
feat: update client tests workflow and dependencies

alacambre-yseop:
In this PR I update the test procedure as described [here](https://code.visualstudio.com/api/working-with-extensions/testing-extension#migrating-from-vscode) and bring back to life the unit tests for the client part.


client/src/test/runTest.ts
```diff
@@ -0,0 +1,22 @@
+import * as path from 'path';
+import { runTests } from 'vscode-test';
+
+async function main() {
+    try {
+        // The folder containing the Extension Manifest package.json
+        // Passed to `--extensionDevelopmentPath`
+        const extensionDevelopmentPath = path.resolve(__dirname, '../../../');
```
mminot-yseop: Lolwut. From `../..` onwards, I expect a comment explaining roughly where the path ends up.

client/src/test/runTest.ts
```diff
@@ -0,0 +1,22 @@
...
+
+        // The path to the extension test runner script
+        // Passed to --extensionTestsPath
+        const extensionTestsPath = path.resolve(__dirname, './suite/index');
```
mminot-yseop: Probably ⇔ and less confusing:

```suggestion
        const extensionTestsPath = path.resolve(__dirname, 'suite/index');
```

Check if you will.

client/src/test/runTest.ts
```diff
@@ -0,0 +1,22 @@
...
+
+        // Download VS Code, unzip it and run the integration test
+        await runTests({ extensionDevelopmentPath, extensionTestsPath });
+    } catch (err) {
+        console.error('Failed to run tests');
```
mminot-yseop: Is it OK to do nothing at all with the caught `err`? It might contain valuable information.

client/src/test/suite/index.ts
```diff
@@ -0,0 +1,44 @@
...
+    ui: 'tdd', // the TDD UI is being used in extension.test.ts (suite, test, etc.)
+    useColors: true, // colored output from test results
+    fullStackTrace: true,
+    reporter: 'xunit',
+    reporterOptions: {
+        // Path from the place the tests are run. In our case: /target/test-reports/client-tests.xml
+        output: '../target/test-reports/client-tests.xml',
+    },
+  });
+  mocha.useColors(true);
```
mminot-yseop: image

client/src/test/suite/index.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const testsRoot = path.resolve(__dirname, '..');
+
+  return new Promise((c, e) => {
```
mminot-yseop: What is `c`?

client/src/test/runTest.ts
mminot-yseop: I know there’s the line “The folder containing the Extension Manifest package.json” but with all this client / server stuff this is still confusing.

client/src/test/suite/index.ts
alacambre-yseop: I have absolutely no clue about this.

client/src/test/suite/index.ts
```diff
@@ -3,42 +3,41 @@ import * as Mocha from 'mocha';
...
+    const mocha = new Mocha({
+        ui: 'tdd', // the TDD UI is being used in extension.test.ts (suite, test, etc.)
+        fullStackTrace: true,
+        reporter: 'xunit',
+        reporterOptions: {
+            // Path from the place the tests are run. In our case: /target/test-reports/client-tests.xml
+            output: '../target/test-reports/client-tests.xml',
+        },
+    });
+    mocha.useColors(true);
```
mminot-yseop: So you removed the `useColors` that was in the constructor call rather than the one that stands on its own after it? Didn’t the opposite work? Doesn’t it make more sense to keep the whole config in one place, namely the constructor call? :thinking:

alacambre-yseop: > So you removed the useColors that was in the constructor call rather than the one that stands on its own after it?

Yes, but no specific reason to do that.

> Didn’t the opposite work?

It did.

> Doesn’t it make more sense to keep the whole config in one place, namely the constructor call?

I didn't think about it at all. To me, because the constructor call and this method call are in the same place, and since both solutions have the exact same effect, I just choose randomly.

mminot-yseop: I think it’s very confusing and gives the impression that you didn’t include it in the initial configuration object thingy for a particular reason.

alacambre-yseop: OK, done.

mminot-yseop:
Testing.

> Downloading VS Code 1.36.0 from https://update.code.visualstudio.com/1.36.0/linux-x64/stable

:hourglass_flowing_sand:

mminot-yseop:
> update#setState idle

This for ¼ h xD

alacambre-yseop:
DAFUK? Even Jenkins has done it correctly.

mminot-yseop:
Jenkins starts with a cleaner repo, I suppose.

PR #109, opened 2019-07-10
build: enable sonarcloud checks

alacambre-yseop:
I'm sorry because this PR is a little bit tricky and has some downsides. Let me explain.
I started to update the Jenkinsfile and add a simple configuration for sonar. This explains the picture in the first comment.
Then I tried to have code coverage too.
The thing is that in this project we were using a test framework named `mocha`. When I searched in the repositories I know, I found `jest`. `jest` provides code coverage very easily with just some nice dev dependencies and has a wonderful integration with VSCode. So, I started to migrate from `mocha` to `jest`.
While everything went fine for the server-side code, client-side tests **don't work anymore**. 
This is because we try to launch tests over an instance of VSCode that will bring its dependencies at a weird moment. Jest tries the code, says something like “`vscode` dependency doesn't exist.”, then fails miserably.
It tried [this method](https://medium.com/@soloydenko/end-to-end-testing-vs-code-extensions-via-jest-828e5edfeb75), but it is based on an older version of VSCode and doesn't work with modern projects.
Every new file or change in the `client/` directory are for that, but I disabled the tests because they are not working (yet).
Server-side tests work fine and code coverage on this part is very nice as you can see:
image
image

While writing this comment, I discovered [this tool](https://istanbul.js.org/) that probably allows to use `mocha`.  while still having code coverage.
I still prefere `jest` now, even for test writing.

Since our client-tests aren't quite good or very relevant, I propose to keep them disabled until I found a way to enable them later. I know that having fewer tests, even dumb ones, isn't nice, but I believe that code coverage and bug smell detection is better.


alacambre-yseop:
I'm progressing!
image

client/src/test/extension.test.ts
```diff
@@ -1,69 +1,61 @@
...
-        assert.notStrictEqual(ymlExtension, null, "Extension doesn't exist.");
+        expect(ymlExtension).not.toBeNull();
         vscode.languages
             .getLanguages()
             .then((languages) => {
-                assert.notStrictEqual(languages, null);
-                assert.notStrictEqual(languages.length, 0);
-                assert.notStrictEqual(languages.indexOf('yml'), -1);
+                expect(languages).not.toBeNull();
+                expect(languages.length).not.toBe(0);
```
mminot-yseop: Nothing like `expect(x).toBeEmpty()`? Just asking xD

client/src/test/index.ts
```diff
@@ -0,0 +1,19 @@
+import { runCLI } from 'jest-cli';
+import path = require('path');
+
+const jestTestRunnerForVSCodeE2E: ITestRunner = {
```
mminot-yseop: E2E?

sonar-project.properties
```diff
@@ -0,0 +1,47 @@
...
+# Files ignored from coverage analysis
+sonar.coverage.exclusions=**/test/*.test.ts
+
+# Load lcov coverage report
+sonar.typescript.lcov.reportPaths=target/coverage/lcov.info
+
+##
+## Duplications
+##
+# Increase minimum lines for code duplications triggering (default = 10)
```
mminot-yseop: Reason for the increase? 20 seems a bit much.

client/src/test/extension.test.ts
Harukero: It seems it [doesn't exist](https://jestjs.io/docs/en/using-matchers). Maybe I can use `is([])` instead?

client/src/test/index.ts
Harukero: E2E → End to End; You can check [this article](https://medium.com/@soloydenko/end-to-end-testing-vs-code-extensions-via-jest-828e5edfeb75) for more information. It's the one from the first comment.

sonar-project.properties
Harukero: No good reason for the increase. Just a copy-paste.

Harukero:
I will try to give a try to https://github.com/microsoft/vscode-test/issues/37.

client/src/test/extension.test.ts
mminot-yseop: Nonono, it’s OK.

sonar-project.properties
mminot-yseop: Try to reset (if not done already) and see if it puts us in dire straits or not. The default value – or even something lower xD – would be nice.

babel.config.js
```diff
@@ -0,0 +1,11 @@
+// babel.config.js
+module.exports = {
+    presets: [
+        ["@babel/preset-env", {
+            targets: {
+                node: "current",
+            },
+        }],
+        +"@babel/preset-typescript",
```
lgregoire-yseop: Isn't the `+` weird, right ?

client/package.json
```diff
@@ -14,23 +14,24 @@
...
     "engines": {
-        "vscode": "^1.34.0"
+        "vscode": "^1.38.0"
     },
     "scripts": {
-        "test": "tsc -b ./tsconfig.json && node ./out/test/runTest.js"
+        "test": "jest"
     },
     "dependencies": {
+        "@types/jest": "^24.0.18",
```
lgregoire-yseop: You can move it to `devDependencies`

client/package.json
```diff
@@ -14,23 +14,24 @@
...
     "devDependencies": {
         "@types/glob": "^7.1.1",
-        "@types/mocha": "^5.2.7",
+        "@types/jest-cli": "^23.6.0",
         "@types/node": "^12.0.12",
-        "@types/vscode": "^1.34.0",
+        "@types/vscode": "^1.38.0",
         "glob": "^7.1.4",
-        "mocha": "^5.2.0",
+        "jest-cli": "^24.9.0",
```
lgregoire-yseop: You can directly use the `jest` dep which simply wraps the jest-cli

client/package.json
```diff
@@ -14,23 +14,24 @@
...
     "dependencies": {
+        "@types/jest": "^24.0.18",
         "event-stream": "3.3.4",
         "typescript": "^3.5.2",
         "vscode-languageclient": "^5.3.0-next.5"
     },
     "devDependencies": {
         "@types/glob": "^7.1.1",
-        "@types/mocha": "^5.2.7",
+        "@types/jest-cli": "^23.6.0",
```
lgregoire-yseop: I suggest to use the `@types/jest` dep to include all jest typings.

server/package.json
```diff
@@ -15,16 +15,16 @@
         "node": "*"
     },
     "dependencies": {
+        "@types/jest": "^24.0.18",
```
lgregoire-yseop: all @types can be moved to devDependencies

sonar-project.properties
lgregoire-yseop: It was a need to allow specific patterns on AA; default values are sufficient I think.

sonar-project.properties
```diff
@@ -0,0 +1,47 @@
...
+sonar.cpd.ts.minimumLines=20
+
+##
+## Lint
+##
+# TSLint report
+# sonar.typescript.tslint.reportPaths=target/lint/tslint-report.json
+
+# Stylelint report
+# sonar.css.stylelint.reportPaths=target/lint/stylelint-report.json
```
lgregoire-yseop: I don't think stylelint will be used one day in this project.

Jenkinsfile
```diff
@@ -23,7 +25,7 @@ pipeline {
                 ansiColor('xterm') {
                     sh "npm install --unsafe-perm"
```
lgregoire-yseop: You could use `npm ci` to install packages without updating them.

Jenkinsfile
```diff
@@ -23,7 +25,7 @@ pipeline {
...
                     sh "npm run compile"
-                    sh "npm run test"
+                    sh "npm run test:ci"
```
lgregoire-yseop: Suggestion: It may be clearer to have these different phases separated.

client/src/test/index.ts
lgregoire-yseop: I don't know what are recommendations for vscode extensions' testing.
This file seems to be the main input to launch tests, but isn't it possible to directly provide the jest command as a TestRunner ?

package.json
```diff
@@ -163,31 +163,36 @@
...
         "antlr4ts-cli": "^0.5.0-alpha.3",
+        "jest": "^24.9.0",
+        "jest-cli": "^24.9.0",
+        "jest-junit": "^8.0.0",
+        "jest-sonar-reporter": "^2.0.0",
         "js-yaml": "^3.13.1",
         "merge-options": "^1.0.1",
-        "mocha": "^5.2.0",
         "rimraf": "^2.6.3",
+        "sonar-scanner": "^3.1.0",
```
lgregoire-yseop: If you use the Sonar CLI provided by Jenkins, this dep is not used.

jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    moduleFileExtensions: ["js", "ts"],
+    preset: "ts-jest",
+    roots: ["<rootDir>/server"],
+    testRegex: "\\.test\\.ts$",
```
mminot-yseop, line 5: Is it normal for this regex to only include one of the `moduleFileExtensions`?

alacambre-yseop: Here is the documentation for `moduleFileExtensions`:
> An array of file extensions your modules use. If you require modules without specifying a file extension, these are the extensions

It also says:
> if you are using TypeScript, you may want to consider moving "ts" and/or "tsx" to the beginning of the array.

→ I have something to do here :sweat_smile: 

For `testRegex`:
> The pattern or patterns Jest uses to detect test files.

It seems to be quite different.

alacambre-yseop:
> Did you fix your issue regarding the config file that seemed to be ignored? (cf. what you said on Slack while asking for help)

It seems so (I hope).

PR #112, opened 2019-07-15
chore: update changelog

alacambre-yseop:


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Release Notes
 
+## 1.6.2 (YYYY-MM-DD)
+
+-   Fix security issue with `lodash` version 4.17.11. New version `4.17.14` set instead.
+-   Display method documentation when hovering method calls.
```
mminot-yseop: It’s not only for calls, is it? Pretty sure I tested that on a method’s declaration itself.
:heavy_plus_sign: ↓

> the definition reads "intransitive verb". So you can hover in a boat, or over the ground. But you cannot hover a boat, nor can you hover the ground.

### =

```suggestion
-   Display a method’s documentation when hovering over its name.
```

alacambre-yseop: > It’s not only for calls, is it? Pretty sure I tested that on a method’s declaration itself.

I tested it only on method calls since this was the thing that bothers me the most.

PR #115, opened 2019-07-17
feat(grammar): multiple parser improvements 

alacambre-yseop:
Sorry for this probably hard to read pull request.
In it I:
- fix some parsing problems
- split Yml.g4 file into a parser file and a lexer file
- thanks to the splitting I can now use two lexer modes − one for the real code and one for the files declarations file (like project.kao)
- reworked the unit tests file to have cleaner test suites
- added unit tests, of course

If possible avoid asking questions with drawings :D


server/src/test/Parsing.test.ts
```diff
@@ -111,22 +134,36 @@ args {
...
             checkInputValidityForRule(
                 (parser) => parser.function(),
                 `
-function Test::applyMyFunction(Function testFunction)
---> action {
-    testFunction.apply();
-}
-;
-`,
+function LibDocument:printRangeValues(LibKPI:KeyPerformanceIndicator KPI,
```
mminot-yseop: Hum… If `LibKPI` and co. are not public yet, consider turning some identifiers into gibberish for tests instead of revealing stuff. Maybe.

server/src/test/Parsing.test.ts
```diff
@@ -163,6 +201,51 @@ complete StudyConcept
...
+        });
+    });
+
+    describe('ymlId rule', () => {
+        it('should accept different kinds of ymlId', (done) => {
+            checkInputValidityForRule((parser) => parser.ymlId(), `abc`);
+            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc`);
+            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc:d`);
+            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc::d`);
+            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc:::d`);
```
mminot-yseop: Would something with strictly more than three colons be accepted? Would it make sense in YML? For the Engine?
If you reject those or end up editing stuff to reject them, consider adding a test for that (provided that there isn’t one already).

server/grammar/YmlLexer.g4
```diff
@@ -0,0 +1,153 @@
...
+
+mode DECLARATION_FILE;
+
+DECL_FILE_PREPROCESSING: PREPROCESSING -> channel(HIDDEN);
+DECL_FILE_LINE_COMMENT: LINE_COMMENT -> channel(HIDDEN);
+DECL_FILE_MULTILINE_COMMENT: MULTILINE_COMMENT -> channel(HIDDEN);
+
+NAME: '.'? ALPHANUM+ ('-' ALPHANUM+)*;
+
+// Could be more complicated, but should be enough in a first place.
```
mminot-yseop: I think you’ve been using this phrase wrong lately.

```suggestion
// Could be more complicated, but should be enough for the time being.
```

image

And regarding “in _the_ first place”, it usually refers to the past.

> _adverb sequence, idiomatic_ To begin with; earlier; first; at the start.

server/grammar/YmlLexer.g4
```diff
@@ -0,0 +1,153 @@
...
+
+PREPROCESSING: '@' ~[\n\r]* -> channel(HIDDEN);
+LINE_COMMENT: '//' ~[\n\r]* -> channel(HIDDEN);
+MULTILINE_COMMENT:
+    MULTILINE_COMMENT_START .*? MULTILINE_COMMENT_END -> channel(HIDDEN)
+;
+
+mode DECLARATION_FILE;
+
+DECL_FILE_PREPROCESSING: PREPROCESSING -> channel(HIDDEN);
```
mminot-yseop: Wouldn’t this deserve one of your “blabla, room for improvement” comments?

server/grammar/YmlLexer.g4
```diff
@@ -0,0 +1,153 @@
...
+DECL_FILE_LINE_COMMENT: LINE_COMMENT -> channel(HIDDEN);
+DECL_FILE_MULTILINE_COMMENT: MULTILINE_COMMENT -> channel(HIDDEN);
+
+NAME: '.'? ALPHANUM+ ('-' ALPHANUM+)*;
+
+// Could be more complicated, but should be enough in a first place.
+FILE_DECLARATION: ('./'? | '../'*) NAME ('/' NAME)* ('.' NAME)*;
```
mminot-yseop: I don’t quite get the purpose of the final `('.' NAME)*` :thinking: Especially since there seems to be another (optional) `.` in `NAME`. Can you show me an example of line that needs this to be recognized?

mminot-yseop: Lol, could be `DECL_FILE_FILE_DECLARATION` but :nauseated_face:

mminot-yseop: What if we just looked for something equivalent to the regex `[^@()]+`? Provided that comments take precedence over this mess.

server/src/test/Parsing.test.ts
```diff
@@ -88,9 +99,21 @@ function LibCube:Fact::getMeasureValue()
             );
             done();
         });
-    });
 
-    describe('function introduced with type Function', () => {
+        it('should parse correctly a function with a Function object as argument', (done) => {
```
mminot-yseop: We say “parse correctly”, but strictly speaking we’re just checking the absence of errors and not the deduced meaning, right? But I think I asked you something very similar ages ago. :face_with_head_bandage: Can’t quite remember the answer.

server/grammar/YmlParser.g4
```diff
@@ -294,10 +171,10 @@ nonArithmeticValue: chainedCall | bool | STRING | DATE;
 
 instanciationVariable: QUESTION_MARK ymlId;
 
+expressionWithIndex: expression withIndex*;
```
mminot-yseop: Huuuum… Confusing “with” in rule name since there can be no index here (`*`). Also, the rule name `withIndex` sounds strange to me too since it’s _just_ an index and we don’t know _what_ is “with” it.

```
possiblyIndexedExpression: expression index*;
```

? Maybe you’ll say it’s ugly as hell.

server/src/test/Parsing.test.ts
```diff
@@ -46,37 +46,48 @@ describe('Parsing Tests', () => {
...
                 (parser) => parser.chainedCall(),
                 `applyCollection(LibCube:Measure, __where currentElement.includeInJointureConditions == true).toList()`,
             );
             done();
         });
-    });
 
-    describe('applyCollection + __operation rule', () => {
-        it('should parse correctly an `applyCollection` instruction with __operation keyword', (done) => {
+        it('should parse correctly a chained call with indexed accesses', (done) => {
```
mminot-yseop: I see only one call here, judging by the number of parentheses pairs. Is that normal? How could it be “chained”, then? :thinking:

server/src/test/Parsing.test.ts
alacambre-yseop: For this file, you're completely right. `should parse without errors` should be better.
There is another file where we check the resulting objects from the parsing.

server/src/test/Parsing.test.ts
alacambre-yseop: Yeah, you're right. I did it for the last tests but didn't update the old ones.

server/src/test/Parsing.test.ts
alacambre-yseop: > Would something with strictly more than three colons be accepted? Would it make sense in YML? For the Engine?

Nothing more should be accepted it doesn't make sense and I guess it would make the compilation fails too.

> If you reject those or end up editing stuff to reject them, consider adding a test for that (provided that there isn’t one already).

I don't have any test proving that an expression that should fail really fails.

server/grammar/YmlLexer.g4
```diff
@@ -0,0 +1,153 @@
...
+
+//We store the whitespace in hidden channel to take back the ones around comments
+WS: [ \r\t\n]+ -> channel(HIDDEN);
+
+DOUBLE: MINUS? NUMBER+ (DOT NUMBER+)?;
+INTEGER: NUMBER+;
+
+/*
+ * Colon is OK inside a YMLID
+ * Simple colon has no specific meaning, however:
```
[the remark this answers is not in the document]
mminot-yseop: Oh, so it’s kinda like when in AWK you use something like `@` in the strings used to index values in an array in order to simulate a multidimensional array even though, behind the scenes, it’s still nothing but a hash map (`0@0`, `0@1`, …, `1@0`, …, `<k>@<k>`)? Hum.

server/grammar/YmlLexer.g4
mminot-yseop: Consecutive dots are OK, you know.

```
    $ find Music/ -type f -wholename '*....*'
[…]
Music/mesarthim/_.-_-..._..._._-._-.-/flac/05_-_......flac
[…]
Music/drudkh/they_often_see_dreams_about_the_spring/flac/02_-_u_dakhiv_irzhavim_kolossyu....flac
[…]
Music/wedrujacy_wiatr/tam_gdzie_miesiac_oplakuje_swit/flac/02_-_tam_gdzie_miesiac_oplakuje_swit....flac
[…]
```

### Trivia

The `_` at the beginning of the `.-_-..._..._._-._-.-` album was manually added to prevent it from becoming a hidden file.

server/src/test/Parsing.test.ts
mminot-yseop: OK for _should parse without errors_.

server/src/test/Parsing.test.ts
mminot-yseop: kay

server/src/test/Parsing.test.ts
mminot-yseop: Is there something preventing you from adding some tests like this?

server/grammar/YmlLexer.g4
alacambre-yseop: `mesarthim` seems great.

server/src/test/Parsing.test.ts
alacambre-yseop: Yes: lack of time.

server/grammar/YmlLexer.g4
alacambre-yseop: Sorry, I think I don't understand what should be the next step on this issue.

mminot-yseop: Well, I think your pattern is – ironically – needlessly strict and complicated.

I might make errors because Antlr noob, but intuitively I’d write:

```
NAME: ('-' | '.' | ALPHANUM)+;
FILE_DECLARATION: NAME ('/' NAME)*;
```

mminot-yseop: Oh and `_` should be accepted too, right?

alacambre-yseop: I had to improve a little your rules, but it was great.
`_` is already in ALPHANUM.

mminot-yseop: > `_` is already in ALPHANUM.

:open_mouth: × 10

server/grammar/YmlLexer.g4
```diff
@@ -0,0 +1,152 @@
...
+
+mode DECLARATION_FILE;
+
+DECL_FILE_PREPROCESSING: PREPROCESSING -> channel(HIDDEN);
+DECL_FILE_LINE_COMMENT: LINE_COMMENT -> channel(HIDDEN);
+DECL_FILE_MULTILINE_COMMENT: MULTILINE_COMMENT -> channel(HIDDEN);
+
+NAME: ('-' | '.' | ALPHANUM)+?;
+// Could be more complicated, but should be enough for the time being.
+FILE_DECLARATION: NAME+ ('/' NAME)*;
```
mminot-yseop: I’ll approve the PR, but here’s a quick note: paths ending with `/.` are accepted and should probably not since this is bound to refer to a directory (does it make sense to include a directory in such a file? Dunno). I think that in the future something like `(n /)* n` instead of `n (/ n)*` would help. Indeed, there can be garbage in the first segments, but we would have to make sure the path does not end with a name that is `.` or `..`. That’s pretty much the only exceptions I can think about. But these are rather overkill considerations.

PR #116, opened 2019-07-17
fix: get documentation for full labels too

alacambre-yseop:
Thank you for your integration work xD

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -69,9 +69,13 @@ connection.onHover((_params) => {
         return null;
     }
     // Multiple elements can have the same shortName
-    const entity = completionProvider.getFirstItemByShortNameMatching(entityName, (elem) => !!elem.documentation);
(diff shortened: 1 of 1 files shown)


mminot-yseop:
I don’t understand this PR. Add explanations while I finish reading _your other_ PR. xd

alacambre-yseop:
1. Let's say that I have a class `MyClass` declaration a method `myMethod`.
2. When implementing this function, I'll have to create a function named `MyClass::myMethod`.
3. When using this method I will call it `myMethod()`.
The shortName of this method is `myMethod` but its full label is `MyClass::myMethod`.

When we hover over the name of the implementation (2) we get as token `MyClass::myMethod`, so we need to look for the full name when searching for the documentation.
When we hover over the name of the method in all other cases (1 and 3), we get as token `myMethod`. So in these cases, the short name is the one we are looking for.

mminot-yseop:
Oh!

Issue #121, opened 2019-07-29
`applyCollectionOn` management


davanessoff-yseop:
Extract:
```
Function SentenceGenerator::getUniqueConcepts
args {}
--> return applyCollectionOn(_dataStructure in this.getDataStructures(), select _dataStructure.filterConcept).toOrderedSet()
;
```
It seems the parser doesn't take this instruction into account.

alacambre-yseop:
It doesn't know the `select` keywork in this context.

davanessoff-yseop:
> It doesn't know the `select` keyword in this context.

I can give you more examples of usage of applyCollectionOn if you want

davanessoff-yseop:
[https://github.com/yseop/yseop-engine/blob/40875809c1297ad07a8d62fbe9a0a1c35d92523c/tests/collection/applyCollection/applyCollection.kao](url)

davanessoff-yseop:
[https://github.com/yseop/yseop-engine/blob/master/tests/collection/applyCollection/applyCollection.kao](url)

PR #122, opened 2019-07-30
feat(grammar): multiple grammar improvements

alacambre-yseop:
- accept `implementation` as a valid ID
- accept `foreach`, `for` and `while` with just one instruction as body
- parse all the cases of `applyCollection` and `applyCollectionOn` (see #121 )


server/src/test/Parsing.test.ts
```diff
@@ -51,14 +51,83 @@ describe('Parsing Tests', () => {
...
                 `applyCollection(myCollection , __operation Namespace:ClassName::member)`,
             );
             done();
         });
+        it('should parse without errors an `applyCollection` instruction with __operation keyword on function name', (done) => {
+            checkInputValidityForRule((parser) => parser.applyCollection(), `applyCollection(list0, __operation carre);`);
+            done();
+        });
+
+        it('should parse without errors an `applyCollection` instruction with __operation ans __arguments keywords', (done) => {
```
mminot-yseop: ?

```suggestion
        it('should parse without errors an `applyCollection` instruction with __operation and __arguments keywords', (done) => {
```

PR #123, opened 2019-07-31
fix: switch instr with case without par; applyCollectionOn as call

alacambre-yseop:


server/src/test/Parsing.test.ts
```diff
@@ -313,6 +313,42 @@ implementation SentenceToGenerate
         });
     });
 
+    describe('switch rule', () => {
+        it('should parse without errors a switch with cases without parenthesis', (done) => {
```
mminot-yseop:
```suggestion
        it('should parse without errors a switch with cases without parentheses around the value', (done) => {
```

server/src/test/Parsing.test.ts
```diff
@@ -313,6 +313,42 @@ implementation SentenceToGenerate
...
+    case VALUE_2 : {
+        // do something else
+    }
+}
+;`,
+            );
+            done();
+        });
+
+        it('should parse without errors a switch with cases between parenthesis', (done) => {
```
mminot-yseop:
```suggestion
        it('should parse without errors a switch with parentheses around values', (done) => {
```

server/src/test/Parsing.test.ts
```diff
@@ -313,6 +313,42 @@ implementation SentenceToGenerate
...
+    }
+    case (VALUE_2) : {
+        // do something else
+    }
+}
+;`,
+            );
+            done();
+        });
+    });
```
mminot-yseop: If the styles can be mixed in a single _switch_, consider adding an example for that.

```suggestion
    });
    
        it('should parse without errors a switch with parentheses around some values', (done) => {
            checkInputValidityForRule(
                (parser) => parser.instruction_switchCase_withValue(),
                `
switch( val ) {
    case (VALUE_1) : {
        // do stuff
    }
    case VALUE_2 : {
        // do something else
    }
}
;`,
            );
            done();
        });
    });
```

PR #124, opened 2019-07-31
fix: more possible hashmap keys; switchExpr and more

### LINKED ISSUE #99: Incorrect error rising at Explicit HashMap
The yml extension marks as an error some hashmaps, for example those where the keys are numbers.
image

alacambre-yseop:
Let's wait first for David's feedback on the tests examples.
Maybe I extended some stuff too much.

This should at least fix #99.

@davanessoff-yseop Do you agree to be more and more called as a reviewer for this kind of PRs?
It can be great to have a pro to check that the tests used are coherent.


davanessoff-yseop:
I can review a bit, but I don't really understand the language. I'm just guessing. I'll focus on the tests.

server/src/test/Parsing.test.ts
```diff
@@ -41,6 +41,35 @@ describe('Parsing Tests', () => {
...
+            checkInputValidityForRule(
+                (parser) => parser.instruction_assignment(),
+                `
+finalVal = switchExpr( myValue ) {
+    case CASE_1 : val1
+    case CASE_2 : val2
+    case CASE_3 : val3
+    case CASE_4 : val4
+};
+`,
```
davanessoff-yseop: there is a confusion here:
`xxxExclusive` only means that if more than one case can be activated an error is produced. It can be used in both switch with or without variable.
All in all, you can have:
`switchExpr`, `switchExpr(val)`, `switchExprExclusive`, `switchExprExclusive(val)`.

Of course, the use of `xxxExclusive` is more relevant with a `switch`  that doesn't have a variable since you could have overlapping conditions easily if you don't pay attention

alacambre-yseop: Nice! It will make the grammar simpler :smile:

alacambre-yseop: @davanessoff-yseop I rewrote the test case, should be better now. Well, it should fail for any number greater than 0.

PR #130, opened 2019-08-09
feat: syntax highlighting using YAML

alacambre-yseop:
In this PR we:
- make possible to use YAML instead of JSON to develop the syntax highlighting
- for now on, `syntaxes/yml.tmLanguage.json` is generated directly from the YAML file
- improve the syntax highlighting for numbers, documentation, inline instances type and probably more.

This PR doesn't fix `Colours odd behavior` yet. It just tries to make it easier to solve later.


CONTRIBUTING.adoc
```diff
@@ -32,11 +32,11 @@ The language server is located in the `server`{nbsp}directory.
 
 ## How to update syntax highlighting
 
-In order to update the syntax highlighting, you must edit `yml.tmLanguage.json` using http://manual.macromates.com/en/language_grammars#language_grammars[the TextMate manual] as reference.
+In order to update the syntax highlighting, you must edit `syntaxes/yml.tmLanguage.YAML-tmLanguage` using http://manual.macromates.com/en/language_grammars#language_grammars[the TextMate manual] as reference. Then you need to convert this file into JSON and replace the content of `syntaxes/yml.tmLanguage.json` which is the file used by the extention for syntax highlighting.
```
mminot-yseop: How is the conversion into JSON to be performed?

CONTRIBUTING.adoc
```diff
@@ -32,11 +32,11 @@ The language server is located in the `server`{nbsp}directory.
...
 
 The format used throughout this manual is slightly different from the one we are using{nbsp}(JSON). It is, however, easy to understand how to adapt examples.
 
-
+NOTE: We use YAML format to develop the syntax highlighting because YAML _allows comments_ in the code and regexp edition is way easier than in JSON because we don't need too much character escape. For more information, check https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#using-yaml-to-write-a-grammar[the VSCode extension manual] about using YAML to write a grammar.
```
mminot-yseop: The combo of “because”-s makes me want a list very badly.

```suggestion
[NOTE]
====
We use YAML format to develop the syntax highlighting because YAML:

* allows comments in the code;
* makes regular expression edition way easier because we don't need to escape many characters.

For more information, check https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#using-yaml-to-write-a-grammar[the VS Code extension manual] about using YAML to write a grammar.
====
```

syntaxes/yml.tmLanguage
```diff
@@ -8,6 +8,14 @@
     <string>Yseop Markup Language</string>
```
mminot-yseop: So there is

* this weird XML-ish version
* the generated JSON one
* the weird one with the overlong extension `YAML-tmLanguage` (gosh, my renaming scripts assume anything longer than “torrent” can’t be an extension)?

I don’t get what this version, here (XML-ish), does or where it comes from.

syntaxes/yml.tmLanguage.YAML-tmLanguage
```diff
@@ -0,0 +1,220 @@
+$schema: 'https://raw.githubusercontent.com/martinring/tmlanguage/master/tmlanguage.json'
+name: 'Yseop Markup Language'
+scopeName: source.yseop
+patterns:
+    -
```
mminot-yseop: Wut? Isn’t there a more concise way to write this inclusion list? I don’t even get what that means, in terms of structures (lists, etc.). Is that a list of singleton maps? Gosh. Sure you can’t provide a list as value for `include`? :open_mouth:

CONTRIBUTING.adoc
alacambre-yseop: It's indicated [here](https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#using-yaml-to-write-a-grammar)

syntaxes/yml.tmLanguage.YAML-tmLanguage
alacambre-yseop: I used a tool to convert the JSON file into this one first. 
It seems that 
```-
    include: '#foo'
```
has the same meaning as
```
- {include: '#foo'}
```

syntaxes/yml.tmLanguage
alacambre-yseop: I guess we can now remove completely this file.

CONTRIBUTING.adoc
mminot-yseop: Then say it xD + “extention” → “extension”

```suggestion
In order to update the syntax highlighting, you must edit `syntaxes/yml.tmLanguage.YAML-tmLanguage` using http://manual.macromates.com/en/language_grammars#language_grammars[the TextMate manual] as reference. Then you need to convert this file into JSON https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide#using-yaml-to-write-a-grammar[as indicated here] and replace the content of `syntaxes/yml.tmLanguage.json` which is the file used by the extension for syntax highlighting.
```

syntaxes/yml.tmLanguage
mminot-yseop: Try and see what happens :+1:

syntaxes/yml.tmLanguage.YAML-tmLanguage
mminot-yseop: I almost tried to write an AWK script to change that stuff but screw that (mostly because of indentation and stuff).
:woman_shrugging:

PR #131, opened 2019-08-30
perf(parsing): improve “parse the whole project” feature

alacambre-yseop:
The goal of this PR is to improve the way we parse the whole project.
Now, instead of parsing every files that have the correct extension, we parse the `project.kao` file and check all of the files it references and then parse all of them. This is done recursively.
We then parse `yclass`, `yobject` and `ycomplete` files.

So, most of the files parsed are now files that are more or less referenced for the project. No more YAML files parsed, for example!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 ## 1.6.5 (YYYY-MM-DD)
 
--   Greatly improved project files parsing. Less useless files are parsed. Now:
+-   Improved project files parsing. Fewer irrelevant files are parsed. The new behavior is as follows:
     -   parse the files mentioned in the `project.kao` file;
     -   parse `*.yclass`, `*.yobject` and `*.ycomplete` files;
-    -   avoid parsing the files in `.generated-yml/` directory, wherever it is.
+    -   avoid parsing the files in `.generated-yml/` directory, wherever this directory stands.
 -   Improved parsing for:
     -   `forall` instructions;
     -   `(condition).check()`-like expressions;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.6.5 (YYYY-MM-DD)
 
+-   Greatly improved project files parsing. Less useless files are parsed. Now:
+    -   parse the files mentioned in the `project.kao` file;
+    -   parse `*.yclass`, `*.yobject` and `*.ycomplete` files;
+    -   avoid parsing the files in `.generated-yml/` directory, wherever it is.
 -   Improved parsing for:
     -   `forall` instructions;
     -   `(condition).check()`-like expressions;
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -170,23 +170,25 @@ function openProjectFile(fileUri: Uri): boolean {
                 .split('\n')
                 // line can be indented in the file.
                 .map((line) => line.trim())
-                /*
-                 * Ignore:
-                 * - any empty line
-                 * - lines that are just preprocessing or Yseop Engine instruction
-                 * - lines that are one-line comments
-                 * - every file from any .generated-yml/ directory
-                 */
                 .filter((line) => {
                     return (
+                        // Ignore empty lines
                         line.length > 0 &&
+                        // Ignore lines that are just preprocessing or Yseop Engine instruction
                         !line.startsWith('@') &&
+                        // Ignore the lines with the _FILE_TYPE_ instruction
                         !line.startsWith('_FILE_TYPE_') &&
+                        // Ignore single-line comments
                         !line.startsWith('//') &&
-                        line.search(/(^\.generated-yml\/)|(\/\.generated-yml\/)/) === -1
+                        // Ignore multi-lines comments starting with “/*”
+                        !line.startsWith('/*') &&
+                        // Ignore multi-lines comments starting with just “*“ (includes “*/”)
+                        !line.startsWith('*') &&
+                        // Drop files from any .generated-yml/ directory
                     );
                 })
                 .map((line) => path.join(path.dirname(doc.uri.fsPath), line))
+                // Make sure the file exists and drop directories
                 .filter((filePath) => fs.existsSync(filePath) && !fs.lstatSync(filePath).isDirectory())
                 .map((filePath) => Uri.parse(`file://${filePath}`))
                 .forEach((uri) => openProjectFile(uri));
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -156,13 +156,18 @@ function openProjectFile(fileUri: Uri): boolean {
         // The document exists and was successfully opened and should be parsed already.
         (doc) => {
             console.debug(`Parsing ${fileUri}.`);
-            const lines = doc.getText().split('\n');
-            if (lines.length === 0 || !lines[0].trim().startsWith('_FILE_TYPE_')) {
+            if (
+                doc
+                    .getText()
+                    .trim()
+                    .startsWith('_FILE_TYPE_')
+            ) {
                 // We are not in a `project.kao`-like file. Do not go further.
                 wasKaoFile = false;
                 return;
             }
-            lines
+            doc.getText()
+                .split('\n')
                 // line can be indented in the file.
                 .map((line) => line.trim())
                 /*
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -122,9 +122,10 @@ export function activate(context: ExtensionContext) {
     // First check recursively all the files set in `project.kao`-like files, starting with `project.kao`.
     workspace.findFiles(`**/project.kao`).then((uris) => {
         uris.forEach((uri) => {
-            openProjectFile(uri);
-            // There should be only one `project.kao` file.
-            return;
+            if (openProjectFile(uri)) {
+                // There should be only one `project.kao` file. If we found a good candidate, stop the loop.
+                return;
+            }
         });
     });
 
@@ -144,8 +145,12 @@ export function activate(context: ExtensionContext) {
  * read its content and apply this function recursivly for each line that are existing file's URI.
  *
  * @param fileUri An existing file URI.
+ *
+ * @return `true` only if the provided URI was a `*.kao`-like file i.e. its first line starts with `_FILE_TYPE_`.
  */
-function openProjectFile(fileUri: Uri): void {
+function openProjectFile(fileUri: Uri): boolean {
+    let wasKaoFile = true;
+
     // Try to open the file. If it is opened, the server will parse it.
     workspace.openTextDocument(fileUri).then(
         // The document exists and was successfully opened and should be parsed already.
@@ -154,6 +159,7 @@ function openProjectFile(fileUri: Uri): void {
             const lines = doc.getText().split('\n');
             if (lines.length === 0 || !lines[0].trim().startsWith('_FILE_TYPE_')) {
                 // We are not in a `project.kao`-like file. Do not go further.
+                wasKaoFile = false;
                 return;
             }
             lines
@@ -186,6 +192,7 @@ function openProjectFile(fileUri: Uri): void {
             }
         },
     );
+    return wasKaoFile;
 }
 
 /**
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -130,8 +130,8 @@ export function activate(context: ExtensionContext) {
     });
 
     /*
-     * List of all the yseopml file extensions known by this extension as set in `client/package.json`
-     * but never explictly set in project.kao-like files.
+     * List of the file-extensions of yseopml files that are never set by the user in `project.kao`-like files.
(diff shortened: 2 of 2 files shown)


client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+                 * Ignore:
+                 * - any empty line
+                 * - lines that are just preprocessing stuff or Yseop Engine instruction.
+                 * - every file from .generated-yml/ directory
+                 */
+                if (
+                    line.length === 0 ||
+                    line.indexOf('@') >= 0 ||
+                    line.indexOf('_FILE_TYPE_') >= 0 ||
+                    line.indexOf('.generated-yml') >= 0
```
mminot-yseop: That’s a bit prone to false positives. If you’ve got time, consider something like

### Skip

```properties
.generated-yml
.generated-yml/
./.generated-yml
foo/.generated-yml
foo/.generated-yml/bar
```

### Parse

```properties
foo/plop.generated-yml
foo/.generated-yml.patate/bar
[…]
```

You’ll probably hate me for suggesting to make something that stupid more robust, though. But to me, this more or less deserves a function of its own rather than an `indexOf`. xD

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+                ) {
+                    return;
+                }
+                const maybeFilePath = path.join(path.dirname(doc.uri.fsPath), line);
```
mminot-yseop: You’re trying to build an absolute path from the relative one found in the file, right? Please rename the variable to make this clear, as it took me like eight seconds to get it.

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+                if (!fs.existsSync(maybeFilePath)) {
+                    // File doesn't exist. Go to the next line.
+                    return;
```
mminot-yseop: I’m pondering whether to warn the dev in such cases… But maybe this would be a job for another part of the program, with, like, red markers in the GUI on the considered line in the file, etc.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.6.5 (YYYY-MM-DD)
 
+-   Greatly improved project files parsing. Less useless files are parsed. Now:
```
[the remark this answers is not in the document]
alacambre-yseop: Well, it used to be poorly implemented. :sweat_smile:

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+ * @param fileUri An existing file URI.
+ */
+function openProjectFile(fileUri: Uri): void {
+    // Try to open the file. If it is opened, the server will parse it.
+    workspace.openTextDocument(fileUri).then(
+        // The document exists and was successfully opened and should be parsed already.
+        (doc) => {
+            console.debug(`Parsing ${fileUri}.`);
+            const lines = doc.getText().split('\n');
+            if (lines.length === 0 || !lines[0].trim().startsWith('_FILE_TYPE_')) {
```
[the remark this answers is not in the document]
alacambre-yseop: I tried but it wasn't required at all for the projects I checked.

client/src/extension.ts
alacambre-yseop: Clearly not. We might encounter comments in the file or some other weird things.
Also, this feature is done without really opening the file, so displaying markers is impossible.
Plus, sometimes there are preprocessor instructions saying “If the file XXX exists […]”. It is not really our goal to do the job of the compiler.

client/src/extension.ts
alacambre-yseop: OK, I'll rename it. But note that the path built can be something that cannot be a real path.

client/src/extension.ts
alacambre-yseop: Well, in our case none of them will be parsed.
And I clearly don't want to handle such false positive right know because of their almost zero probability in YML projects.

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+ *
+ * @param fileUri An existing file URI.
+ */
+function openProjectFile(fileUri: Uri): void {
+    // Try to open the file. If it is opened, the server will parse it.
+    workspace.openTextDocument(fileUri).then(
+        // The document exists and was successfully opened and should be parsed already.
+        (doc) => {
+            console.debug(`Parsing ${fileUri}.`);
+            const lines = doc.getText().split('\n');
```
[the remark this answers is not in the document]
alacambre-yseop: I'll try to improve this in this PR.

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+            if (lines.length === 0 || !lines[0].trim().startsWith('_FILE_TYPE_')) {
+                // We are not in a `project.kao`-like file. Stop the going further.
+                return;
```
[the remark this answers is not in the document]
alacambre-yseop: I can do that in another PR, if you don't mind.

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+            }
+            lines.forEach((line) => {
+                // line can be indented in the file.
+                line = line.trim();
```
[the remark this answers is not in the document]
alacambre-yseop: I'll do this in this PR.

CHANGELOG.md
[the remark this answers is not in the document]
mminot-yseop: That’s not something to brag about!

client/src/extension.ts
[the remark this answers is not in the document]
mminot-yseop: ? Would it be that hard? You already have `return` statements. Why not just add `false` and a `if` where it’s called?

client/src/extension.ts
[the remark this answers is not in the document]
mminot-yseop: :woman_shrugging:

client/src/extension.ts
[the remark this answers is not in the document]
alacambre-yseop: Not hard. Just time-consuming.

client/src/extension.ts
mminot-yseop: > in our case none of them will be parsed

That’s the point, hehe.

> almost zero probability

That’s the kind of answer I expected, but the current implem still strikes me as an incredibly sloppy way to handle paths. With a proper object type, you could loop through the path segments and find whether one of them is `.generated-yml` in a flash.

client/src/extension.ts
mminot-yseop: Urg. If you know about weird cases, mention them in a code-comment. This sounds fishy.

client/src/extension.ts
mminot-yseop: Huuum… If we check file existences for each comment line, it might end up being _more costly_ than the previous implem, don’t you think? Couldn’t we add dumb filtering conditions to weed out at least some comments? You did so for preprocessing instructions.

client/src/extension.ts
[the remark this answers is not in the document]
mminot-yseop: :open_mouth: Really? Seemed pre-chewed to me.

client/src/extension.ts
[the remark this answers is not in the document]
alacambre-yseop: I couldn't avoid the line storing because of the check done line 160.

client/src/extension.ts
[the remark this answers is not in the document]
alacambre-yseop: Should be done now.

client/src/extension.ts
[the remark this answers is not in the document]
alacambre-yseop: Should be done now.

client/src/extension.ts
```diff
@@ -116,16 +118,82 @@ export function activate(context: ExtensionContext) {
...
+
+                /*
+                 * Ignore:
+                 * - any empty line
+                 * - lines that are just preprocessing stuff or Yseop Engine instruction.
+                 * - every file from .generated-yml/ directory
+                 */
+                if (
+                    line.length === 0 ||
```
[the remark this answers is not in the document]
alacambre-yseop: Should be done now.

client/src/extension.ts
alacambre-yseop: Done more or less now. I didn't find an object type to handle path segments in JS/TS.

client/src/extension.ts
alacambre-yseop: I added the one-line comments as filtered case.

client/src/extension.ts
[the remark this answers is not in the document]
mminot-yseop: Which check? I think stuff moved and I don’t see anything relevant xD

client/src/extension.ts
```diff
@@ -116,16 +118,85 @@ export function activate(context: ExtensionContext) {
...
+                // We are not in a `project.kao`-like file. Do not go further.
+                wasKaoFile = false;
+                return;
+            }
+            lines
+                // line can be indented in the file.
+                .map((line) => line.trim())
+                /*
+                 * Ignore:
+                 * - any empty line
```
mminot-yseop: * Again, I’d move each of these comments next to the corresponding test.
* I’d use a succession of filters rather than a single obese one.
* Aren’t you recompiling the same regex over and over again, here?
* As far as I know, `/* …` is more common than `//` in those files. Wouldn’t hurt to skip that too. Sure, we won’t be able to tell how many lines the comment spans, but at least we will be skipping the first one of each (which is often the only one, as in the typical `/** end of project specific beforeExternReferences.dcl **/` lines).

```ts
// Empty lines.
.filter((line) => line.length > 0)
// Preprocessing instructions.
.filter((line) => !line.startsWith('@'))
// File type header.
.filter((line) => !line.startsWith('_FILE_TYPE_'))
// Comments.
.filter((line) => !line.startsWith('//'))
.filter((line) => !line.startsWith('/*'))
// Files in “generated YML” directories.
.filter((line) => line.search(BLAHBLAH_CONSTANT_PRE_COMPILED_REGEX)/) === -1)
```

(Did what I could with my limited knowledge xD)

client/src/extension.ts
```diff
@@ -116,16 +118,85 @@ export function activate(context: ExtensionContext) {
...
+                    return (
+                        line.length > 0 &&
+                        !line.startsWith('@') &&
+                        !line.startsWith('_FILE_TYPE_') &&
+                        !line.startsWith('//') &&
+                        line.search(/(^\.generated-yml\/)|(\/\.generated-yml\/)/) === -1
+                    );
+                })
+                .map((line) => path.join(path.dirname(doc.uri.fsPath), line))
+                .filter((filePath) => fs.existsSync(filePath) && !fs.lstatSync(filePath).isDirectory())
```
mminot-yseop:
```suggestion
                // Make sure the file exists.
                .filter((filePath) => fs.existsSync(filePath))
                // Drop directories.
                .filter((filePath) => !fs.lstatSync(filePath).isDirectory())
```

client/src/extension.ts
alacambre-yseop: > * Again, I’d move each of these comments next to the corresponding test.

OK!

> * I’d use a succession of filters rather than a single obese one.

In fact, each call to `filter` does the following: `Returns the elements of an array that meet the condition specified in a callback function.`. I don't want to have a new array created for each small piece of condition. Don't forget that we don't have here a compilator that will increase code performance. There is no JAVA-like `Stream` neither.

> * Aren’t you recompiling the same regex over and over again, here?

There is a high probability here, yeah. Good catch.

> * As far as I know, /* … is more common than // in those files. Wouldn’t hurt to skip that too. Sure, we won’t be able to tell how many lines the comment spans, but at least we will be skipping the first one of each (which is often the only one, as in the typical /** end of project specific beforeExternReferences.dcl **/ lines).

Well, I never see `the typical […] lines` in any of my projects.
But if you say it's that important, I'll add these tests.

mminot-yseop: OK if you think the multiple filters will decrease perfs.

> Well, I never see the typical […] lines in any of my projects.
> But if you say it's that important, I'll add these tests.

```bash
$ locate -0 project.kao | xargs -0 grep -ho '^\s*/[/*]' | sed 's/^\s*//' | sort | uniq -c
      4 //
    410 /*
```

I think those numbers of occurrences speak for themselves.

client/src/extension.ts
```diff
@@ -23,6 +24,7 @@ let yseopCliStatusBarItem: StatusBarItem;
 let yseopCliPath: string;
 let parseAllProjectFilesAtStartup: boolean;
 
+const GENERATED_YML_DIR_REGEX = /(^\.generated-yml\/)|(\/\.generated-yml\/)/;
```
mminot-yseop:
```suggestion
const GENERATED_YML_DIR_REGEX = /(^|\/)\.generated-yml\//;
```

PR #132, opened 2019-08-30
feat: remove option; add diagnostic severity option

alacambre-yseop:


package.json
```diff
@@ -72,6 +66,18 @@
...
+                    "type": "string",
+                    "scope": "window",
+                    "enum": [
+                        "Error",
+                        "Warning",
+                        "Information",
+                        "Hint"
+                    ],
+                    "default": "Information",
+                    "description": "Issues found by parser are marked with a Diagnostic Severity Level. This affects the color of squiggle."
```
mminot-yseop: I assume that, just like in Eclipse, the type of highlighting can be configured, so it’s better, I think, not to explicitly refer to “squiggles” or anything.

```suggestion
                    "description": "Issues found by the parser are marked with a Diagnostic Severity Level. This affects the way these issues are highlighted in the code."
```

server/src/server.ts
```diff
@@ -97,24 +108,27 @@ interface ISettings {
...
 }
 
 let engineModel: EngineModel;
-// hold the maxNumberOfProblems setting
 let pathToPredefinedObjectsXml: string;
 let activateParsingProblemsReporting: boolean;
-// The settings have changed. Is send on server activation
-// as well.
+
+// By default, the severity is Information.
```
mminot-yseop: It surprises me to find this default-value-setting in the middle of nowhere. Is it possible to set it directly in the option’s definition, in the `package.json` file?
Also, maybe we should document this default value (tell which value is the default one) in the option’s description or… I dunno.

server/src/server.ts
```diff
@@ -97,24 +108,27 @@ interface ISettings {
...
+export let severity: DiagnosticSeverity = DiagnosticSeverity.Information;
+
+// The settings have changed. Is send on server activation as well.
 connection.onDidChangeConfiguration((change) => {
     const settings = change.settings as ISettings;
     activateParsingProblemsReporting = settings.yseopml.activateParsingProblemsReporting;
     pathToPredefinedObjectsXml = settings.yseopml.pathToPredefinedObjectsXml;
+    // One of the diagnostic level possible, or Information.
```
mminot-yseop: You’ll have a bit of search-and-replace to do, I’m afraid. And the option’s name itself to fix.

> Diagnostic:
>
> * _adjective_ Of, relating to, or used in a diagnosis.
> * _adjective_ Serving to identify a particular disease; characteristic.
> * _noun_ The art or practice of medical diagnosis.
> * _noun_ A symptom or a distinguishing feature serving as supporting evidence in a diagnosis.
> * _noun_ An instrument or technique used in medical diagnosis.

# Edit

Wut, `DiagnosticSeverity` is not from you? Gosh. Then I dunno. Still, the option’s name is rather vague: it’s not the diagnosis of _everything_ but of parsing errors and the like, right? As soon as we’ll have something entirely different _X_ to diagnose, the option’s name will become confusing if the said option does not affect the level for _X_. But maybe I misunderstand that stuff.

mminot-yseop:
Tested.
image
Consider slipping a “YML” somewhere in the description to make the option easier to find. I had to look for “diagnostic” (and ended up with quite a lot of unrelated stuff).

> Issues found by the parser are marked with a Diagnostic Severity Level. This affects the way these issues are highlighted in the code.

↓↓↓↓↓↓↓↓↓↓↓↓↓↓

> Issues found by the YML parser are marked with a Diagnostic Severity Level. This affects the way these issues are highlighted in the code.

:woman_shrugging:

server/src/server.ts
alacambre-yseop: I'll check if that's possible.

server/src/server.ts
alacambre-yseop: I'll rename the variable to `parsingIssueSeverityLevel`, something like that.

PR #134, opened 2019-10-08
feat: get documentation if any; avoid missing ident case in tag

alacambre-yseop:
PR for future devs on binarisation of `lib/` directory.
It implies to have a `predefinedObjects.xml` which handles the `lib/` definitions instead of parsing this directory.
It implies that we need to get an access to the doc from this XML file instead of the code itself.

### DIFF
--- a/server/src/engineModel/EngineModel.ts
+++ b/server/src/engineModel/EngineModel.ts
@@ -33,7 +33,9 @@ export class EngineModel {
         if (!fs.existsSync(this.uri)) {
             connection.console.warn(`File “${this.uri}” doesn't exist.`);
         } else if (!fs.lstatSync(this.uri).isFile()) {
-            connection.console.warn(`It seems that ${this.uri} is not a normal file. Therefore, it cannot be imported.`);
+            connection.console.warn(
+                `It seems that ${this.uri} is not a normal file. Therefore, it cannot be imported.`,
+            );
         } else {
             this.parsePredefinedObjects(this.uri);
         }
@@ -58,6 +60,9 @@ export class EngineModel {
             ymlAbstractFunction.detail = `[STATIC] ${ymlAbstractFunction.label}`;
         }
         ymlAbstractFunction.data = `id_${sourceType ? sourceType : 'static'}_${ymlAbstractFunction.label}`;
+        if (!!func.doc) {
+            ymlAbstractFunction.documentation = func.doc[0];
+        }
         return ymlAbstractFunction;
     }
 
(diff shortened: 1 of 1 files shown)


server/src/engineModel/EngineModel.ts
```diff
@@ -58,6 +60,9 @@ export class EngineModel {
             ymlAbstractFunction.detail = `[STATIC] ${ymlAbstractFunction.label}`;
         }
         ymlAbstractFunction.data = `id_${sourceType ? sourceType : 'static'}_${ymlAbstractFunction.label}`;
+        if (!!func.doc) {
+            ymlAbstractFunction.documentation = func.doc[0];
```
mminot-yseop: Why is `doc` an array, again? :thinking: Is it OK to take just `[0]`?

server/src/engineModel/EngineModel.ts
```diff
@@ -227,8 +238,12 @@ export class EngineModel {
     private importTags(dataAndFeatures: any): void {
         // Get the single “text-tags” element.
         const textTags = dataAndFeatures['text-tags'][0];
-        textTags.tag.forEach((tag) => {
-            const textTag = this.buildYmlFunction(tag);
+        textTags.tag.forEach((_tag) => {
+            if (!_tag.$.ident) {
```
mminot-yseop: Can’t remember what those tags are and what is happening here :| sorry.
Could you please refresh my memory with examples or explanations or whatever?

server/src/engineModel/EngineModel.ts
alacambre-yseop: Here, `func` is something like an XML node with children. Each child may be present multiple time or just once.
```
<func>
  <myChild/>
  <myChild/>
</func>
```
That's why it is an array. 
We can't have more than one `doc` element in our case under `func`, so it is OK to take just `[0]`.

server/src/engineModel/EngineModel.ts
alacambre-yseop: Tags in YML are the elements that are like functions, but starting with `\`. They are used in `TextGranule` and `TextFunction` objects.

mminot-yseop: And what about `.$.ident`?

alacambre-yseop: `.$` is the object containing the tag's attributes. `ident` is one of them (the YML identifier).

mminot-yseop: How can I test that stuff?

PR #136, opened 2019-10-18
refactor: move code and rework the way the code is split

alacambre-yseop:
Restructured the code to split each components responsibility.
The work here is greatly inspired by [this repo](https://github.com/freebroccolo/ocaml-language-server/tree/master/src/bin/server). 
There is still plenty of work to do but I think the current state is enough for now.

### DIFF
--- a/server/src/feature/didChangeConfiguration.ts
+++ b/server/src/feature/didChangeConfiguration.ts
@@ -1,0 +1,39 @@
+import { DiagnosticSeverity, DidChangeConfigurationParams, NotificationHandler } from 'vscode-languageserver-protocol';
+
+import { EngineModel } from '../engineModel/EngineModel';
+import Session from '../session';
+
+/**
+ * Map between the string available as option and real `DiagnosticSeverity` enum values.
+ */
+const diagSeverityMap = new Map<string, DiagnosticSeverity>([
+    ['error', DiagnosticSeverity.Error],
+    ['warning', DiagnosticSeverity.Warning],
+    ['information', DiagnosticSeverity.Information],
+    ['hint', DiagnosticSeverity.Hint],
+]);
+
+let engineModel: EngineModel;
+let pathToPredefinedObjectsXml: string;
+let activateParsingProblemsReporting: boolean;
+// By default, the severity is Information.
+export let parsingIssueSeverityLevel: DiagnosticSeverity = DiagnosticSeverity.Information;
+
+export default function(session: Session): NotificationHandler<DidChangeConfigurationParams> {
+    return (change) => {
+        session.onDidChangeConfiguration(change);
+        activateParsingProblemsReporting = change.settings.yseopml.activateParsingProblemsReporting;
+        pathToPredefinedObjectsXml = change.settings.yseopml.pathToPredefinedObjectsXml;
+        // One of the severity levels possible, or `Information`.
+        parsingIssueSeverityLevel =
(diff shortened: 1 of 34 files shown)


server/src/session/index.ts
```diff
@@ -0,0 +1,70 @@
...
+    IPCMessageWriter,
+    TextDocumentChangeEvent,
+} from 'vscode-languageserver-protocol';
+
+import { YmlCompletionItemsProvider } from '../completion/YmlCompletionItemsProvider';
+import { YmlDefinitionProvider } from '../definitions';
+import * as validateTextDocument from '../feature';
+import { ISettings } from '../interfaces';
+
+export default class Session implements Disposable {
```
mminot-yseop: I admit I don’t quite get this concept of “session”. For example, what is a session’s scope? Is there one session per window? Etc.
Consider adding a few words of documentation for dummies like me.
It seems also slightly strange to me to have direct access through that class to the `warning`, etc. logging methods. I don’t see sessions printing stuff that often. :thinking: Oh, that seems to be what this OCaml project does. But what made you hold that project in such high esteems, to begin with? :thinking:

server/src/session/index.ts
```diff
@@ -0,0 +1,70 @@
...
+        // Make the text document manager listen on the connection
+        // for open, change and close text document events
+        this.documents.listen(this.connection);
+        // Listen on the connection
+        this.connection.listen();
+    }
+
+    public async onDidChangeConfiguration({ settings }: DidChangeConfigurationParams): Promise<void> {
+        (this.settings as any) = { ...this.settings, ...settings };
+    }
```
mminot-yseop: My feeble knowledge of TS does not allow me to understand such things. :neutral_face:

server/src/feature/didChangeConfiguration.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default function(session: Session): NotificationHandler<DidChangeConfigurationParams> {
+    return (change) => {
+        session.onDidChangeConfiguration(change);
+        activateParsingProblemsReporting = change.settings.yseopml.activateParsingProblemsReporting;
+        pathToPredefinedObjectsXml = change.settings.yseopml.pathToPredefinedObjectsXml;
+        // One of the severity levels possible, or `Information`.
+        parsingIssueSeverityLevel =
+            diagSeverityMap.get(change.settings.yseopml.ymlParsingIssueSeverityLevel.toLowerCase()) ||
+            DiagnosticSeverity.Information;
```
mminot-yseop: The default value is kinda written in three places, now, right? Here, in the `export let … = DiagnosticSeverity.Information;`, and in the constructor of `YmlParsingErrorListener`. Seems strange.

Btw, I saw no call of that constructor (`YmlParsingErrorListener`) that was making use of the optional second argument meant to pass the severity level. I don’t quite get how this is handled now.

server/src/feature/validateTextDocument.ts
```diff
@@ -0,0 +1,42 @@
...
+        const textDocUri = textDocument.uri;
+        const diagnostics: Diagnostic[] = [];
+
+        // Create the lexer and parser
+        const inputStream = CharStreams.fromString(textDocument.getText());
+        const lexer = new YmlLexer(inputStream);
+        const tokenStream = new CommonTokenStream(lexer);
+        const parser = new YmlParser(tokenStream);
+        parser.removeErrorListeners();
+        parser.addErrorListener(new YmlParsingErrorListener(diagnostics));
```
mminot-yseop: Maybe naive question: do we recreate and redo _all this_ at each change of a file? :open_mouth:

PR #138, opened 2019-11-11
feat(defs): use go to implementation feature

Harukero:
In this PR I apply the two suggestions brought by @mminot-yseop about the use of `pop()` and the comment.

> So it can be declared in several places? Hum… D’you mean, like, not technically the same method, but homonyms throughout the workspace, for example? Or is this a YML feature consisting of declaring the same symbol multiple times?

Also, I try to answer his concerns about the fact that we are mixing declaration and implementation. It should be clearer now, even though it is not perfect yet.

I also fix a security issue reported by `npm audit`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
     -   `forall` instructions;
     -   `(condition).check()`-like expressions;
     -   `rename` instructions.
--   Improved definition provider. `Go To Definition` on a method name gives the list of the locations where the method is declared and where it is implemented.
+-   Improved location provider. `Go To Implementation` using `ctrl+F12` on a method name gives the list of the locations where the method is is implemented. See [this link](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation) for more information.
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -12,7 +12,7 @@
     -   `forall` instructions;
     -   `(condition).check()`-like expressions;
     -   `rename` instructions.
--   Improved definition provider. `Go To Definition` on a method name gives the list of the locations where the method is declared and where it is implemented.
+-   Improved location provider. `Go To Implementation` using `ctrl+F12` on a method name gives the list of the locations where the method is is implemented. See [this link](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation) for more information.
```
alacambre-yseop:
```suggestion
-   Improved location provider. `Go To Implementation` using `ctrl+F12` on a method name gives the list of the locations where the method is implemented. See [this link](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation) for more information.
```

PR #141, opened 2019-11-22
feat(parse): parse project client-side; avoid leak warning

alacambre-yseop:
In this PR we move the whole parsing work to the server-side, avoiding to do as many calls to the server as there are files to parse.
This was making the extension crash sometimes because of potential memory leaks. This PR fixes this issue too.

**NOTE**: I'll do a release after this one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,17 @@
 
 -   Remove unused option `maxNumberOfProblems`;
 -   New option `ymlParsingIssueSeverityLevel` allowing to set the way parsing problems are displayed;
--   Improved project files parsing. Fewer irrelevant files are parsed. The new behavior is as follows:
-    -   parse the files mentioned in the `project.kao` file;
-    -   parse `*.yclass`, `*.yobject` and `*.ycomplete` files;
-    -   avoid parsing the files in `.generated-yml/` directory, wherever this directory stands.
 -   Improved parsing for:
     -   `forall` instructions;
     -   `(condition).check()`-like expressions;
     -   `rename` instructions.
 -   Improved location provider. `Go To Implementation` using `ctrl+F12` on a method name gives the list of the locations where the method is implemented. See [this link](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation) for more information.
+-   Improved project files parsing. Fewer irrelevant files are parsed. The new behavior is as follows:
+    -   parse the files mentioned in the `project.kao` file;
+    -   parse `*.yclass`, `*.yobject` and `*.ycomplete` files;
+    -   avoid parsing the files in `.generated-yml/` directory, wherever this directory stands.
+
+_Note that the whole project parsing may take some time, depending on your project size. During the project parsing, suggestions list won't be available._
 
 ## 1.6.4 (2019-07-31)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
     -   parse `*.yclass`, `*.yobject` and `*.ycomplete` files;
     -   avoid parsing the files in `.generated-yml/` directory, wherever this directory stands.
 
-_Note that the whole project parsing may take some time, depending on your project size. During the project parsing, suggestions list won't be available._
+_Note that parsing a whole project may take some time, depending on the project’s size. During the project parsing, suggestion lists won't be available._
 
 ## 1.6.4 (2019-07-31)
 
--- a/package.json
+++ b/package.json
@@ -182,21 +182,21 @@
     },
     "devDependencies": {
         "@types/mocha": "^5.2.7",
-        "@types/node": "^11.13.15",
+        "@types/node": "^11.15.3",
         "antlr4ts-cli": "^0.5.0-alpha.3",
         "js-yaml": "^3.13.1",
         "merge-options": "^1.0.1",
         "mocha": "^5.2.0",
-        "rimraf": "^2.6.3",
+        "rimraf": "^2.7.1",
         "ts-loader": "^5.4.5",
-        "tslint": "^5.18.0",
+        "tslint": "^5.20.1",
         "tslint-config-prettier": "^1.18.0",
-        "tslint-consistent-codestyle": "^1.15.1",
+        "tslint-consistent-codestyle": "^1.16.0",
         "tslint-eslint-rules": "^5.4.0",
         "tslint-sonarts": "^1.9.0",
-        "typescript": "^3.5.2",
-        "vsce": "^1.69.0",
-        "webpack": "^4.35.0",
-        "webpack-cli": "^3.3.5"
+        "typescript": "^3.7.3",
+        "vsce": "^1.70.0",
+        "webpack": "^4.41.2",
+        "webpack-cli": "^3.3.10"
     }
 }
--- a/package.json
+++ b/package.json
@@ -195,7 +195,7 @@
         "tslint-eslint-rules": "^5.4.0",
         "tslint-sonarts": "^1.9.0",
         "typescript": "^3.5.2",
-        "vsce": "^1.64.0",
+        "vsce": "^1.69.0",
         "webpack": "^4.35.0",
         "webpack-cli": "^3.3.5"
     }
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -4,6 +4,10 @@
  * ------------------------------------------------------------------------------------------ */
 'use strict';
 import { CharStreams, CommonTokenStream } from 'antlr4ts';
+import * as fs from 'fs';
+import * as glob from 'glob';
+import * as path from 'path';
+import * as url from 'url';
 import {
     CompletionItem,
     createConnection,
@@ -13,7 +17,6 @@ import {
     InitializeResult,
     IPCMessageReader,
     IPCMessageWriter,
-    MarkupKind,
     TextDocument,
     TextDocumentChangeEvent,
     TextDocumentPositionParams,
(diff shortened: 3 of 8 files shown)


mminot-yseop:
I’d be interested to see a time comparison (from launch to “completion suggestions are available”) but you probably have better things to do xD

alacambre-yseop:
> I’d be interested to see a time comparison (from launch to “completion suggestions are available”) but you probably have better things to do xD

In fact, it doesn't bring performance improvements. However, it avoids the `LEAK` warnings and some crashes because of that.
Also, since everything is done in the server thread, it should help avoid client freeze.

server/src/server.ts
```diff
@@ -74,6 +78,127 @@ connection.onInitialize(
...
+ * @param extension The extension of the files to look for
+ */
+function parseFilesWithExtension(workspacePath: string, extension: string): void {
+    glob(`**/*.${extension}`, {
+        cwd: workspacePath,
+        ignore: '**/.generated-yml/**',
+    })
+        .then((_matches) => {
+            if (!_matches) {
+                return;
```
mminot-yseop: 1) I’m surprised to see there are nice `cwd` and `ignore` parameters to these nice globbing tools but no way to immediately discard items that do not match.
2) Btw, are we sure we’re not going to try to open directories as if they were regular files? Did you try creating weird dirs like `foo.<extension>/`?
3) Btw-2, could it be optimized by calling it only once, with multiple extensions? I dunno if these tools allow patterns like `foo.{yml,yclass,yobject}`. That would for example factorize the `ignore: …` checks. But maybe this would not be significant. I dunno. I tried to see how you were using that function (e.g., if you were calling it multiple times in consecutive lines or if these calls were spread out more broadly) but it seems to be either in collapsed areas of the diff or even outside of it. :( If you like this idea and think it could be applicable, you can always create a GitHub ticket for it or something, I guess.

server/src/server.ts
```diff
@@ -74,6 +78,127 @@ connection.onInitialize(
...
+                fsPromises
+                    .readFile(uri)
+                    .then((_doc) => {
+                        const doc = _doc.toString();
+                        parseFile(`file://${workspacePath}/${uri}`, doc);
+                    })
+                    .catch((_err) => {
+                        if (!!_err) {
+                            connection.console.error(`${_err}`);
+                            return;
```
mminot-yseop: I see the `return` is in the `then`. Is it intended, like “abort for all files as soon as one file makes things go wrong”?
Btw, how likely are we to `catch` an error that would be falsey? And is it better to ignore falsey errors (provided that such things exist) than to log them anyway? This `if` seems strange.

Gosh, sometimes those reviews sound more like you’re teaching me TS than proper reviews.

server/src/server.ts
```diff
@@ -74,6 +78,127 @@ connection.onInitialize(
...
+                        }
+                    });
+            });
+        })
+        .catch((_error) => console.error(_error));
+}
+
+const GENERATED_YML_DIR_REGEX = /(^|\/)\.generated-yml\//;
```
mminot-yseop: Stranded const amidst functions and whatnot? :thinking:

server/src/server.ts
```diff
@@ -140,12 +265,16 @@ connection.onDidChangeConfiguration((change) => {
...
 
+function parseFile(textDocUri: string, docContent: string) {
+    connection.console.log(`Yseop.vscode-yseopml − Parsing ${textDocUri}`);
+    const diagnostics: Diagnostic[] = [];
     // Create the lexer and parser
-    const inputStream = CharStreams.fromString(textDocument.getText());
+    const inputStream = CharStreams.fromString(docContent, textDocUri);
     const lexer = new YmlLexer(inputStream);
     const tokenStream = new CommonTokenStream(lexer);
     const parser = new YmlParser(tokenStream);
```
mminot-yseop: Do we need these for every file type?
Hum I suppose most of our files are YML anyway so the gain would be negligible.

server/src/server.ts
```diff
@@ -140,12 +265,16 @@ connection.onDidChangeConfiguration((change) => {
 });
 
 function validateTextDocument(textDocument: TextDocument): void {
```
mminot-yseop: If you asked me out of the blue, I would have said that “validation” was part of “parsing” rather than the opposite. Consider adding doc for dumb people like me that have it backwards.

server/src/server.ts
alacambre-yseop: > Stranded const amidst functions and whatnot?

I only understand “functions” and “and” from this comment.

mminot-yseop: C’mon! :pouting_woman: What I mean is that your code goes “weird stuff, then a function, then a constant, then a function again”.

server/src/server.ts
alacambre-yseop: 1. It is completely related to one of your other comments. I wrongly supposed that `[]` was falsy. I'll remove this.
2. Not sure. I'll add a check in the forEach.
3. If `foo.{yml,yclass,yobject}` is a correct glob expression, it should be OK and very helpful indeed. I use this function only during the server initialization and loop over the desired extensions.

server/src/server.ts
alacambre-yseop: > I see the return is in the then. Is it intended, like “abort for all files as soon as one file makes things go wrong”?

The `return` is in the `catch`, but I will remove it, it is useless. This is what remains of a former implementation.

> Btw, how likely are we to catch an error that would be falsey? And is it better to ignore falsey errors (provided that such things exist) than to log them anyway? This if seems strange.

I don't think we can get a falsy error here, but since we can get `any` here I have to do the check, I think.
I can still log the fact that an unknown error occurred, what do you think?

> Gosh, sometimes those reviews sound more like you’re teaching me TS than proper reviews.

Well, I'm a YML dev, not a TS dev. I believe we are both teaching to each other. :)

server/src/server.ts
alacambre-yseop: OK, I'll put related things together.

server/src/server.ts
alacambre-yseop: I can rename it and document it. OK.

server/src/server.ts
alacambre-yseop: > Do we need these for every file type?

Unfortunately, yes. These objects are all required to perform the parsing. The parsing will be done only for certain types of files, so it should be OK to leave it that way. This is controlled by the extension itself.

server/src/server.ts
mminot-yseop: Yeah we could at least add in the `console.error` call a couple of words about the context (`File parsing: …` or whatever; I don’t recall the exact context we had here) and make sure we don’t use the auto `${…}` toString-ish thing on an empty string, a zero, a NaN… xD

mminot-yseop:
I think the loading time (before suggestions are available) is so long that it would deserve a note in the doc or something. To make sure beople understand that it’s not caused by a faulty installation on their side.

alacambre-yseop:
> I think the loading time (before suggestions are available) is so long that it would deserve a note in the doc or something. To make sure beople understand that it’s not caused by a faulty installation on their side.

You're right. Note that the loading time greatly depends on the project's size.

mminot-yseop:
> > I think the loading time (before suggestions are available) is so long that it would deserve a note in the doc or something. To make sure beople understand that it’s not caused by a faulty installation on their side.
> 
> You're right. Note that the loading time greatly depends on the project's size.

Oh, yeah, we could mention that, too. If there’s already something about the “parse all” option, this could be around that place. I don’t really remember what form our documentation takes.

package.json
```diff
@@ -195,7 +195,7 @@
         "tslint-eslint-rules": "^5.4.0",
         "tslint-sonarts": "^1.9.0",
         "typescript": "^3.5.2",
-        "vsce": "^1.64.0",
+        "vsce": "^1.69.0",
```
mminot-yseop: 1.70 was released since this edit xD

mminot-yseop:
Seems OK during tests.

server/src/server.ts
```diff
@@ -26,6 +30,24 @@ import { EngineModel } from './engineModel/EngineModel';
...
+    activateParsingProblemsReporting: boolean;
+    pathToPredefinedObjectsXml: string;
+    ymlParsingIssueSeverityLevel: string;
+    parseAllProjectFilesAtStartup: boolean;
+}
+
+let engineModel: EngineModel;
+let pathToPredefinedObjectsXml: string;
+let activateParsingProblemsReporting: boolean;
+let parseAllProjectFilesAtStartup: boolean;
```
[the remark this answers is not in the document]
alacambre-yseop: > I don’t really get why there are these duplications.

The interface just indicates what is the expected format of the object containing the settings that we will handle server-side while the `let`-s just allow to keep track of the value when initialized or changed without having to send a request to the client to get the info.

> Why isn’t ymlParsingIssueSeverityLevel duplicated as well? :thinking:

I guess I miss-used the way we should handle the settings. I will do some fix in another no-that-important PR.

> The relative order of activateParsingProblemsReporting and pathToPredefinedObjectsXml is inconsistent. ^^

I will reorder them, then.

CHANGELOG.md
```diff
@@ -4,15 +4,17 @@
...
     -   `forall` instructions;
     -   `(condition).check()`-like expressions;
     -   `rename` instructions.
 -   Improved location provider. `Go To Implementation` using `ctrl+F12` on a method name gives the list of the locations where the method is implemented. See [this link](https://code.visualstudio.com/docs/editor/editingevolved#_go-to-implementation) for more information.
+-   Improved project files parsing. Fewer irrelevant files are parsed. The new behavior is as follows:
+    -   parse the files mentioned in the `project.kao` file;
+    -   parse `*.yclass`, `*.yobject` and `*.ycomplete` files;
+    -   avoid parsing the files in `.generated-yml/` directory, wherever this directory stands.
+
+_Note that the whole project parsing may take some time, depending on your project size. During the project parsing, suggestions list won't be available._
```
mminot-yseop: I thought for a while that we could add something in the ReadMe but did not find a suitable place for it. :/

Anyway:

```suggestion
_Note that parsing a whole project may take some time, depending on the project’s size. During the project parsing, suggestion lists won't be available._
```

And after that I think we’re good to go.

PR #142, opened 2019-11-22
feat(enum): suggests enum values from predef objects

alacambre-yseop:


server/src/engineModel/EngineModel.ts
```diff
@@ -92,6 +93,15 @@ export class EngineModel {
                 this.completionProvider.addCompletionItem(method);
             });
         }
+
+        if (!!yclass.instances && !!yclass.instances[0].item) {
```
mminot-yseop: 1) This condition could be made clearer via a function.
2) From my point of view, you seem to be assuming that empty arrays are falsey, but according to what I can see on the Web they are truthy, so you can theoretically reach the `yclass.instances[0]` with `instances` holding no element. And if you end up having to add checks for this, that leads us back to (1). ^^
3) Wow, every commit on this extension seems to add logic into which we use `[0]` and ignore the rest. That’s always confusing me. Is it easy to explain with a comment, here?

server/src/engineModel/EngineModel.ts
```diff
@@ -92,6 +93,15 @@ export class EngineModel {
...
+            yclass.instances[0].item.forEach((_instance) => {
```
mminot-yseop: So we’ve got an array of instances, and each one contains an “item” field containing… several instances? It sounds oddly bidimensional, but that’s probably me missing something again.

server/src/engineModel/EngineModel.ts
```diff
@@ -92,6 +93,15 @@ export class EngineModel {
...
+                if (ymlClass.extends.indexOf('yseop.lang.Enum') >= 0) {
```
mminot-yseop: # Style

I see people doing “≠ -1” more often than “≥ 0”. There’s this notion of “index of, not found ⇒ -1” in nearly every Java / JS dev’s mind, I think. But I might be wrong. But the `≥` makes me want to double-check that you used 0 and not something else (like you did for `::` in another PR because you wanted to have at least a non-colon character before the `::`). With ≠ there’s less ambiguity as to what you’re trying to do: it’s kinda clear that the actual position doesn’t matter.

LOL the more trifling the matter is, the longer my comments get.

server/src/engineModel/EngineModel.ts
```diff
@@ -100,6 +110,17 @@ export class EngineModel {
         this.completionProvider.addCompletionItem(ymlClass);
     }
 
+    private buildEnumValue(instanceXmlElement: any, sourceType: string): YmlObjectInstance {
+        const identifier = `${sourceType}::${instanceXmlElement.$.ident}`;
+        const instance = new YmlObjectInstance(identifier, this.uri);
+        instance.detail = `${identifier} − Value of enum ${sourceType}`;
```
mminot-yseop: Any way to get rid of the `Value of enum` (while keeping `${identifier} − ${sourceType}`) in favor of an icon change in the suggestion list or something? Dunno which possibilities are made available to you here.

server/src/engineModel/EngineModel.ts
alacambre-yseop: I'll use `!= -1`. Thanks.

server/src/engineModel/EngineModel.ts
alacambre-yseop: CompletionItemKind has Enum and EnumMember as possible values. I'll try to use them.

server/src/engineModel/EngineModel.ts
```diff
@@ -92,6 +93,27 @@ export class EngineModel {
...
         }
+
+        /* A class (or enum) with instances has the following format:
+        <class ident="EnumClassName">
+            <extends>yseop.lang.Enum</extends>
+            <instances count="2">
+                <item ident="EnumInstance_1"></item>
+                <item ident="EnumInstance_1"></item>
+            </instances>
+        </class>
```
mminot-yseop: Lol, your comment makes it even more strange. I’d expect the `[0]` to directly provide access to the first item.

server/src/engineModel/EngineModel.ts
```diff
@@ -92,6 +93,27 @@ export class EngineModel {
...
+        */
+        if (!!yclass.instances && !!yclass.instances[0].item) {
+            // The current class has already existing instances.
+            // Iterate over each instances items.
```
mminot-yseop: ? I thought that as soon as you write `.item` you only have one instance. What those “items” actually are is obscure to me.

server/src/engineModel/EngineModel.ts
```diff
@@ -92,6 +93,27 @@ export class EngineModel {
...
+            yclass.instances[0].item.forEach((_instance) => {
+                const instance =
+                    ymlClass.extends.indexOf('yseop.lang.Enum') !== -1
```
mminot-yseop: So it can appear anywhere? What if someone defines something like

`foo.bar.yseop.lang.Enum`

or even

`fooyseop.lang.Enumbar`

?

alacambre-yseop: `extends` is an Array of string.

server/src/engineModel/EngineModel.ts
alacambre-yseop: `.item` is the array of `item` tags.

server/src/engineModel/EngineModel.ts
alacambre-yseop: `yclass.instances` is the array of all `instances` tags, even if there is just one element.
`yclass.instances[0].item` is the array of all `item` tags in the first `instances` tag.

server/src/engineModel/EngineModel.ts
mminot-yseop: Omg I somehow missed the “.extends” part and thought that was the class name itself you were rumaging through. Sorry.

server/src/engineModel/EngineModel.ts
mminot-yseop: And why do we focus so much on the first instance, again?
I feel like I’ve asked that in the past but I can’t seem to recall anything so I’ll keep asking for more comments until I get it xD

alacambre-yseop: We focus on the first element in `instances` because there is only one xD

mminot-yseop: Gosh, like, we can’t prevent the parser or whatever from creating an array but in the end it places all instances within a single `instances` element? This framework or whatever it is seems so weird.

mminot-yseop:
LOL the weird hexadecimal suggestions.

codes

mminot-yseop:
Can you give me enum instance examples? I’m baffled by what `grep -iC 5 enum predefinedObjects.xml` gives me.

alacambre-yseop:
> LOL the weird hexadecimal suggestions.

If I find from where they're coming, I will filter them.

> Can you give me enum instance examples? I’m baffled by what `grep -iC 5 enum predefinedObjects.xml` gives me.

You can try this one: `PERIOD_FISCAL_YEAR` or `LEFT_RIGHT`, for example.

alacambre-yseop:
I won't put this one in the next release. Not stable enough.

alacambre-yseop:
I need to rework it. I close it for the moment.

PR #145, opened 2019-12-20
feat: differentiate filetype F and M for kao files

alacambre-yseop:
In a `M` type `*.kao` file, paths are relative to the current *.kao file’s directory.
In a `F` type `*.kao` file, paths are relative to the project's root.
In this PR we assume that the root is the workspace path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## X.X.X (YYYY-MM-DD)
 
--   Fix `*.kao` files parsing at startup.
+-   Fix `*.kao` files parsing at startup (“F” and “M” types differentiation).
 
 ## 1.6.5 (2019-12-19)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## X.X.X (YYYY-MM-DD)
+
+-   Fix `*.kao` files parsing at startup.
+
 ## 1.6.5 (2019-12-19)
 
 -   Remove unused option `maxNumberOfProblems`;
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -62,9 +62,9 @@ const diagSeverityMap = new Map<string, DiagnosticSeverity>([
 /** Regex that matches paths containing the `.generated-yml` directory. */
 const GENERATED_YML_DIR_REGEX = /(^|\/)\.generated-yml\//;
 /** Regex that matches the `_FILE_TYPE_ F` instruction. */
-const FILE_TYPE_F = /^_FILE_TYPE_\s+F\s*/;
+const FILE_TYPE_F = /^_FILE_TYPE_\s+F\b/;
 /** Regex that matches the `_FILE_TYPE_ M` instruction. */
-const FILE_TYPE_M = /^_FILE_TYPE_\s+M\s*/;
+const FILE_TYPE_M = /^_FILE_TYPE_\s+M\b/;
 
 // Create a connection for the server. The connection uses Node's IPC as a transport
 export const connection: IConnection = createConnection(new IPCMessageReader(process), new IPCMessageWriter(process));
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -213,14 +213,13 @@ function openProjectFile(workspacePath: string, fileUri: string): boolean {
                         line.search(GENERATED_YML_DIR_REGEX) === -1
                     );
                 })
-                .map((line) =>
-                    isTypeM
-                        ? // In a M type *.kao file, paths are relative to the current *.kao file.
-                          path.join(path.dirname(fileUri), line)
-                        : // In a F type *.kao file, paths are relative to the project's root.
-                          // Here we assume that the root is the workspace path.
-                          path.join(workspacePath, line),
-                )
+                .map((line) => {
+                    // In a M type *.kao file, paths are relative to the current *.kao file.
+                    // In a F type *.kao file, paths are relative to the project's root.
+                    // Here we assume that the root is the workspace path.
+                    const from = isTypeM ? path.dirname(fileUri) : workspacePath;
+                    return path.join(from, line);
+                })
                 // Make sure the file exists and drop directories
                 .filter((filePath) => fs.existsSync(filePath) && !fs.lstatSync(filePath).isDirectory())
                 .forEach((uri) => openProjectFile(workspacePath, uri));
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -213,15 +213,14 @@ function openProjectFile(workspacePath: string, fileUri: string): boolean {
                         line.search(GENERATED_YML_DIR_REGEX) === -1
                     );
                 })
-                .map((line) => {
-                    if (isTypeM) {
-                        // In a M type *.kao file, paths are relative to the current *.kao file.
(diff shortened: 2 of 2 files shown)


server/src/server.ts
```diff
@@ -207,7 +213,15 @@ function openProjectFile(workspacePath: string, fileUri: string): boolean {
                         line.search(GENERATED_YML_DIR_REGEX) === -1
                     );
                 })
-                .map((line) => path.join(path.dirname(fileUri), line))
+                .map((line) => {
+                    if (isTypeM) {
+                        // In a M type *.kao file, paths are relative to the current *.kao file.
```
mminot-yseop: AH

server/src/server.ts
```diff
@@ -207,7 +213,15 @@ function openProjectFile(workspacePath: string, fileUri: string): boolean {
...
+                        return path.join(path.dirname(fileUri), line);
+                    }
+                    // In a F type *.kao file, paths are relative to the project's root.
+                    // Here we assume that the root is the workspace path.
+                    return path.join(workspacePath, line);
```
mminot-yseop: Might be slightly nicer with an `else` to highlight the M / F opposition.

server/src/server.ts
```diff
@@ -179,8 +183,10 @@ function openProjectFile(workspacePath: string, fileUri: string): boolean {
         .readFile(fileUri)
         .then((_file) => {
             const fileContent = _file.toString();
-            parseFile(`file://${workspacePath}/${fileUri}`, fileContent);
-            if (!fileContent.trim().startsWith('_FILE_TYPE_')) {
+            const isTypeF = fileContent.search(FILE_TYPE_F) !== -1;
```
mminot-yseop: I’m a bit worried about the fact that a “startsWith” turned into a double full regex search. No way to restrain that a bit? Our parsing step is already costly.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## X.X.X (YYYY-MM-DD)
+
+-   Fix `*.kao` files parsing at startup.
```
mminot-yseop:
```suggestion
-   Fix `*.kao` files parsing at startup (“F” and “M” types differentiation).
```

server/src/server.ts
alacambre-yseop: Well, I thought about it a little bit too. But:
* since we cannot be sure of how many spaces there could be between `_FILE_TYPE_ and `F` or `M`, I thought it would be better with a regex. 
* note that the second one is done only when the first one failed.
* in the precedent version, we were doing a `trim` first, maybe this is costly too?
* the pattern we are looking for is only at the start of the string, I suppose it will fail very fast if possible and not be that costly?

server/src/server.ts
alacambre-yseop: OK but it doesn't seem that much of an improvement since there are just a few lines of code + comments.

server/src/server.ts
mminot-yseop: Use numbered lists to make it easier to answer you xD

. Spaces: Sure.
. Only if failed: Yeah but there’s a large majority of files are neither F or M, right? Oh but maybe we only have .kao files, here? Or not?
. Trim: No, I think trim is trivial.
. `^`: I missed that. That makes the other points kinda irrelevant. OK for this stuff, I suppose.

server/src/server.ts
mminot-yseop: In one case you’ve got a clear “this or that”, and in the other (the current version) it looks like F is a garbage-collecting case or something xD.

mminot-yseop: Besides, those comments about the ≠ between F and M would have made a nice PR description :cry:

alacambre-yseop: My linter didn't let me add this useless `else`.
I used the `cond ? a : b;` operator instead.

server/src/server.ts
alacambre-yseop: 2. You're right, damn :/

server/src/server.ts
```diff
@@ -207,7 +213,14 @@ function openProjectFile(workspacePath: string, fileUri: string): boolean {
...
                     );
                 })
-                .map((line) => path.join(path.dirname(fileUri), line))
+                .map((line) =>
+                    isTypeM
+                        ? // In a M type *.kao file, paths are relative to the current *.kao file.
+                          path.join(path.dirname(fileUri), line)
+                        : // In a F type *.kao file, paths are relative to the project's root.
+                          // Here we assume that the root is the workspace path.
+                          path.join(workspacePath, line),
```
mminot-yseop: # If you wish

Duuuude, the `path.join` and the `line` are the same. Consider doing something like…

```ts
from = isTypeM ? path.dirname(fileUri) : workspacePath;
path.join(from, line);
```

Yeah, I’m annoying, but highlighting what changes and what does not changes (+ having a variable name to help) seems nice ^^
Well, actually, I think I’m more disturbed by the fact that your ternary operator spans _five_ freaking lines xD

server/src/server.ts
```diff
@@ -61,6 +61,10 @@ const diagSeverityMap = new Map<string, DiagnosticSeverity>([
 
 /** Regex that matches paths containing the `.generated-yml` directory. */
 const GENERATED_YML_DIR_REGEX = /(^|\/)\.generated-yml\//;
+/** Regex that matches the `_FILE_TYPE_ F` instruction. */
+const FILE_TYPE_F = /^_FILE_TYPE_\s+F\s*/;
+/** Regex that matches the `_FILE_TYPE_ M` instruction. */
+const FILE_TYPE_M = /^_FILE_TYPE_\s+M\s*/;
```
mminot-yseop: Consider (for both of these regexps) replacing the final `\s*` with a `\b`. It expresses what we’re actually looking for slightly better. Here you have some sort of weird optional-ish fake lookahead xD

server/src/server.ts
mminot-yseop: LOL What was the explanation? And that’s called the “ternary” operator xd

mminot-yseop:
I wonder what F and M stand for. Fabsolute and Mrelative? xD

I’ll create a fake KB with two ≠ kaos to test.

mminot-yseop:
Seems OK.

```
Yseop.vscode-yseopml − Parsing file:///home/mminot/bordel/temp/fake_kb/project.kao
Yseop.vscode-yseopml − Parsing file:///home/mminot/bordel/temp/fake_kb/yes-root.yml
Yseop.vscode-yseopml − Parsing file:///home/mminot/bordel/temp/fake_kb/a_dir/yes-dir.yml
Yseop.vscode-yseopml − Parsing file:///home/mminot/bordel/temp/fake_kb/a_dir/sub-project.kao
Yseop.vscode-yseopml − Parsing file:///home/mminot/bordel/temp/fake_kb/a_dir/other_dir/yes.yml
```

```
├── a_dir
│   ├── no-dir.yml
│   ├── other_dir
│   │   ├── no.yml
│   │   └── yes.yml
│   ├── sub-project.kao
│   └── yes-dir.yml
├── no-root.yml
├── project.kao
└── yes-root.yml
```

project.kao

```
_FILE_TYPE_ F

yes-root.yml

a_dir/sub-project.kao

a_dir/yes-dir.yml
```

a_dir/sub-project.kao

```
_FILE_TYPE_ M

other_dir/yes.yml
```

PR #156, opened 2020-04-17
feat(hover): better hover messages

alacambre-yseop:
In this PR I change how we handle hover messages as well as some documentation provided by the completion list.
I thought it was sad to have no documentation or `Not Documented` when hovering elements while we can at least provide elements signature.

I tried to find a format that's not too awful.
It allows to have that kind of elements to be displayed.

```
(variable) indicatorProperty_concept ⇒ NlgApplication||NlgLambdaTerm
(method) [LibDocument:ConceptsWriter].buildDefaultReferentialExpression ⇒ NlgApplication,NlgLambdaTerm
(instance) [GLOBAL] LibOntology:DerivedTree:MC2 ⇒ ConstList
(function) [STATIC] logWarning
(method) [Collection].get
(class) HashSet extends yseop.collection.Collection
```

This detailed part is displayed when hovering elements or in the suggestion list.
When possible, the documentation is displayed to, as usual.
I'm not very proud of how I managed to get a better format. There is probably a better way to do it, but I can't figure it now.
If possible, I would like to do it in another PR.

To do so, I had to add some attributes to `AbstractYmlObject`. This made me see that some of our tests were not very robust to changes and very hard to maintain. I changed these tests' content to make them fit better the intent of what we wanted to test.

~~Last but not least, I finally succeeded to not have bullshit in suggest list by not creating suggestion elements when a file has parsing errors.
Errors are still displayed, but the suggest list doesn't contain entries like `==myElement`.~~

**NOTE**: nothing here fixes the fact that when there are some identifiers duplicated the hover feature will select one more or less randomly.

### DIFF
--- a/server/src/engineModel/EngineModel.ts
+++ b/server/src/engineModel/EngineModel.ts
@@ -52,17 +52,18 @@ export class EngineModel {
     private buildYmlFunction(func: any, sourceType?: string): AbstractYmlFunction {
         let ymlAbstractFunction: AbstractYmlFunction;
         if (sourceType) {
-            ymlAbstractFunction = new YmlFunction(func.$.ident, this.uri);
+            ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
             this.functions.push(ymlAbstractFunction);
             ymlAbstractFunction.detail = `[${sourceType}].${ymlAbstractFunction.label}`;
         } else {
-            ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
+            ymlAbstractFunction = new YmlFunction(func.$.ident, this.uri);
             ymlAbstractFunction.detail = `[STATIC] ${ymlAbstractFunction.label}`;
         }
+        ymlAbstractFunction.setUserInformations(
+            `(${sourceType ? 'method' : 'function'}) ${ymlAbstractFunction.detail}`,
+            !!func.doc ? func.doc[0] : null,
+        );
         ymlAbstractFunction.data = `id_${sourceType ? sourceType : 'static'}_${ymlAbstractFunction.label}`;
-        if (!!func.doc) {
-            ymlAbstractFunction.setDocumentation(func.doc[0]);
-        }
         return ymlAbstractFunction;
     }
 
@@ -92,9 +93,7 @@ export class EngineModel {
                 this.completionProvider.addCompletionItem(method);
             });
         }
-        if (!!yclass.doc) {
-            ymlClass.setDocumentation(yclass.doc[0]);
-        }
+        ymlClass.setUserInformations(`(class) ${ymlClass.label}`, !!yclass.doc ? yclass.doc[0] : null);
         this.classes.push(ymlClass);
         this.enrichYmlClass(ymlClass);
         this.completionProvider.addCompletionItem(ymlClass);
@@ -103,9 +102,10 @@ export class EngineModel {
     private buildAttribute(attributeXmlElement: any, sourceType: string): YmlAttribute {
         const attribute = new YmlAttribute(attributeXmlElement.$.ident, this.uri);
         attribute.detail = `[${sourceType}].${attribute.label}`;
-        if (!!attributeXmlElement.doc) {
-            attribute.setDocumentation(attributeXmlElement.doc[0]);
-        }
+        attribute.setUserInformations(
+            `(property) ${attribute.detail}`,
+            !!attributeXmlElement.doc ? attributeXmlElement.doc[0] : null,
+        );
         attributeXmlElement.return.forEach((returnType) => {
             if (returnType.domains) {
                 attribute.domains = returnType.domains;
@@ -256,6 +256,6 @@ export class EngineModel {
     private enrichYmlClass(yclass: YmlClass): void {
         yclass.data = `id_${yclass.label}`;
         yclass.detail =
-            yclass.extends.length > 0 ? `Class: ${yclass.label} extends ${yclass.extends}` : `Class: ${yclass.label}`;
+            yclass.extends.length > 0 ? `(class) ${yclass.label} extends ${yclass.extends}` : `(class) ${yclass.label}`;
     }
 }
--- a/server/src/engineModel/EngineModel.ts
+++ b/server/src/engineModel/EngineModel.ts
@@ -51,18 +51,18 @@ export class EngineModel {
     }
     private buildYmlFunction(func: any, sourceType?: string): AbstractYmlFunction {
         let ymlAbstractFunction: AbstractYmlFunction;
+        const documentation: string = !!func.doc ? func.doc[0] : null;
         if (sourceType) {
             ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
             this.functions.push(ymlAbstractFunction);
-            ymlAbstractFunction.detail = `[${sourceType}].${ymlAbstractFunction.label}`;
+            ymlAbstractFunction.setUserInformations(
+                `(function) [${sourceType}].${ymlAbstractFunction.label}`,
+                documentation,
+            );
         } else {
             ymlAbstractFunction = new YmlFunction(func.$.ident, this.uri);
-            ymlAbstractFunction.detail = `[STATIC] ${ymlAbstractFunction.label}`;
+            ymlAbstractFunction.setUserInformations(`(function) [STATIC] ${ymlAbstractFunction.label}`, documentation);
         }
-        ymlAbstractFunction.setUserInformations(
-            `(${sourceType ? 'method' : 'function'}) ${ymlAbstractFunction.detail}`,
-            !!func.doc ? func.doc[0] : null,
-        );
         ymlAbstractFunction.data = `id_${sourceType ? sourceType : 'static'}_${ymlAbstractFunction.label}`;
         return ymlAbstractFunction;
     }
@@ -93,7 +93,8 @@ export class EngineModel {
                 this.completionProvider.addCompletionItem(method);
             });
         }
-        ymlClass.setUserInformations(`(class) ${ymlClass.label}`, !!yclass.doc ? yclass.doc[0] : null);
+        const documentation: string = !!yclass.doc ? yclass.doc[0] : null;
+        ymlClass.setUserInformations(`(class) ${ymlClass.label}`, documentation);
         this.classes.push(ymlClass);
         this.enrichYmlClass(ymlClass);
         this.completionProvider.addCompletionItem(ymlClass);
@@ -101,11 +102,8 @@ export class EngineModel {
 
     private buildAttribute(attributeXmlElement: any, sourceType: string): YmlAttribute {
         const attribute = new YmlAttribute(attributeXmlElement.$.ident, this.uri);
-        attribute.detail = `[${sourceType}].${attribute.label}`;
-        attribute.setUserInformations(
-            `(property) ${attribute.detail}`,
-            !!attributeXmlElement.doc ? attributeXmlElement.doc[0] : null,
-        );
+        const documentation: string = !!attributeXmlElement.doc ? attributeXmlElement.doc[0] : null;
+        attribute.setUserInformations(`(property) [${sourceType}].${attribute.label}`, documentation);
         attributeXmlElement.return.forEach((returnType) => {
             if (returnType.domains) {
                 attribute.domains = returnType.domains;
@@ -255,7 +253,9 @@ export class EngineModel {
      */
     private enrichYmlClass(yclass: YmlClass): void {
         yclass.data = `id_${yclass.label}`;
-        yclass.detail =
-            yclass.extends.length > 0 ? `(class) ${yclass.label} extends ${yclass.extends}` : `(class) ${yclass.label}`;
+        yclass.detail = `(class) ${yclass.label}`;
+        if (yclass.extends.length > 0) {
+            yclass.detail = yclass.detail.concat(` extends ${yclass.extends}`);
+        }
     }
 }
--- a/server/src/engineModel/EngineModel.ts
+++ b/server/src/engineModel/EngineModel.ts
@@ -51,7 +51,7 @@ export class EngineModel {
     }
     private buildYmlFunction(func: any, sourceType?: string): AbstractYmlFunction {
         let ymlAbstractFunction: AbstractYmlFunction;
-        const documentation: string = !!func.doc ? func.doc[0] : null;
+        const documentation: string = getDocValue(func);
         if (sourceType) {
             ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
             this.functions.push(ymlAbstractFunction);
@@ -93,7 +93,7 @@ export class EngineModel {
                 this.completionProvider.addCompletionItem(method);
             });
         }
-        const documentation: string = !!yclass.doc ? yclass.doc[0] : null;
+        const documentation: string = getDocValue(yclass);
         ymlClass.setUserInformations(`(class) ${ymlClass.label}`, documentation);
         this.classes.push(ymlClass);
         this.enrichYmlClass(ymlClass);
@@ -102,7 +102,7 @@ export class EngineModel {
 
     private buildAttribute(attributeXmlElement: any, sourceType: string): YmlAttribute {
         const attribute = new YmlAttribute(attributeXmlElement.$.ident, this.uri);
-        const documentation: string = !!attributeXmlElement.doc ? attributeXmlElement.doc[0] : null;
+        const documentation: string = getDocValue(attributeXmlElement);
         attribute.setUserInformations(`(property) [${sourceType}].${attribute.label}`, documentation);
         attributeXmlElement.return.forEach((returnType) => {
             if (returnType.domains) {
@@ -262,4 +262,28 @@ export class EngineModel {
             yclass.detail = yclass.detail.concat(` extends ${yclass.extends}`);
         }
     }
+}
+
+/**
+ * Returns the value of first xmlElement's child named `doc` tag if any.
+ *
+ * E.g. for the following XML:
+ * ```xml
+ * <xmlElement>
+ *     <doc>documentation</doc>
+ * </xmlElement>
+ * ```
+ * its representation is:
+ * ```JSON
+ * {
+ *     doc: ["documentation"]
+ * }
+ * ```
+ *
+ * @param xmlElement a JSON representation of an XML tag.
+ *
+ * @return the xmlElement's documentation or `null`.
+ */
+function getDocValue(xmlElement: any): string {
+    return !!xmlElement.doc ? xmlElement.doc[0] : null;
 }
--- a/server/src/engineModel/EngineModel.ts
+++ b/server/src/engineModel/EngineModel.ts
@@ -265,7 +265,7 @@ export class EngineModel {
 }
 
 /**
- * Returns the value of first xmlElement's child named `doc` tag if any.
+ * Returns the value of the first child `doc` element within a given XML element, if any.
  *
  * E.g. for the following XML:
  * ```xml
--- a/server/src/engineModel/EngineModel.ts
+++ b/server/src/engineModel/EngineModel.ts
@@ -118,25 +118,29 @@ export class EngineModel {
     private parsePredefinedObjects(uri: string): void {
         connection.console.log(`Parsing definition file: ${uri}`);
         fs.readFile(uri, (err, data) => {
-            parser.parseString(data, (parseErr, predefinedObjects) => {
-                if (err != null) {
-                    connection.console.error(`Something went wrong during YE model import:\n ${parseErr}`);
-                } else if (predefinedObjects == null) {
-                    connection.console.error('Something went wrong during YE model import. Your file seems empty.');
-                } else {
-                    try {
-                        const dataAndFeatures = predefinedObjects['data-and-features'];
-                        this.importClasses(dataAndFeatures);
-                        this.importFunctions(dataAndFeatures);
-                        this.importTags(dataAndFeatures);
-                    } catch (importErr) {
-                        connection.console.error(`Something went wrong during YE model import:\n ${importErr}`);
-                    }
+            this.parsePredefinedObjectsFileContent(err, data);
+        });
+    }
+
+    public parsePredefinedObjectsFileContent(err, data: Buffer | string): void {
+        parser.parseString(data, (parseErr, predefinedObjects) => {
+            if (err != null) {
+                connection.console.error(`Something went wrong during YE model import:\n ${parseErr}`);
+            } else if (predefinedObjects == null) {
+                connection.console.error('Something went wrong during YE model import. Your file seems empty.');
+            } else {
+                try {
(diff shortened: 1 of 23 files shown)


server/src/server.ts
```diff
@@ -382,17 +382,21 @@ function parseFile(textDocUri: string, docContent: string) {
     // Reset all the document's completion items.
     completionProvider.removeDocumentCompletionItems(textDocUri);
 
+    // There are syntax errors. Send the errors, if required, to the client and don't visit the failing files.
```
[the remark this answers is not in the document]
alacambre-yseop: In fact, the process was:
1. parse the file and generate error messages to display, if any and required;
1. visit the parsing result and create completion items, whatever the result of the parsing.

Now I propose:

1. parse the file and generate error messages to display, if any and required;
1. if no parsing error, visit the parsing result and create completion items.

If you think this will break too many things, I can remove this commit and postpone another solution.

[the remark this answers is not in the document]
mminot-yseop: ? So if, in a long file, there’s a tiny mistake (or some exotic YML feature that your grammar cannot yet understand), the ser won’t get any suggestion from this file? :thinking:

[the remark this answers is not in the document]
alacambre-yseop: Hum… yeah.
You're right, I remove this commit and will think about a better solution (I have some leads).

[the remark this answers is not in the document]
alacambre-yseop: Removed :+1:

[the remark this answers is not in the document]
mfioravanti-yseop: Great :-) Thanks @mminot-yseop for sharing my opinion :-)

server/src/engineModel/EngineModel.ts
```diff
@@ -52,17 +52,18 @@ export class EngineModel {
...
+            ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
             this.functions.push(ymlAbstractFunction);
             ymlAbstractFunction.detail = `[${sourceType}].${ymlAbstractFunction.label}`;
         } else {
-            ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
+            ymlAbstractFunction = new YmlFunction(func.$.ident, this.uri);
             ymlAbstractFunction.detail = `[STATIC] ${ymlAbstractFunction.label}`;
         }
+        ymlAbstractFunction.setUserInformations(
+            `(${sourceType ? 'method' : 'function'}) ${ymlAbstractFunction.detail}`,
```
mminot-yseop, line 63: * Do people care much about that function vs. method destinction?
* Is it already possible to tell whether it’s a func or meth with just what ends up in `ymlAbstractFunction.detail`?

server/src/engineModel/EngineModel.ts
```diff
@@ -92,9 +93,7 @@ export class EngineModel {
                 this.completionProvider.addCompletionItem(method);
             });
         }
-        if (!!yclass.doc) {
-            ymlClass.setDocumentation(yclass.doc[0]);
-        }
+        ymlClass.setUserInformations(`(class) ${ymlClass.label}`, !!yclass.doc ? yclass.doc[0] : null);
```
mminot-yseop, line 96: Kinda same thing here: maybe the details already show plainly enough that this is a class (but maybe not (I have yet to try it out)).

server/src/engineModel/EngineModel.ts
```diff
@@ -256,6 +256,6 @@ export class EngineModel {
     private enrichYmlClass(yclass: YmlClass): void {
         yclass.data = `id_${yclass.label}`;
         yclass.detail =
-            yclass.extends.length > 0 ? `Class: ${yclass.label} extends ${yclass.extends}` : `Class: ${yclass.label}`;
+            yclass.extends.length > 0 ? `(class) ${yclass.label} extends ${yclass.extends}` : `(class) ${yclass.label}`;
```
mminot-yseop, line 259: Huge-ass ternary thingy with `(class) ${yclass.label}` on both sides. You could yank that part out of there to make it clearer what exactly changes according to `yclass.extends.length`. Kinda old, though, but you placed that stuff right under my nose. :p

```ts
yclass.detail = `(class) ${yclass.label}`
if(yclass.extends.length > 0) {
    yclass.detail += ` extends ${yclass.extends}`
}
```

LOL it’s only while writing this that I understood what this was doing exactly (I wasn’t paying much attention).

server/src/test/YmlCompletionItemsProvider.test.ts
```diff
@@ -51,31 +51,30 @@ describe('YmlCompletionItemsProvider', () => {
...
+        const allDocumentedItemsByLabel = completionProvider.getAllItemsByLabelMatching(methodNameGetName, (elem) =>
+            elem.hasDocumentation(),
         );
         const allDocumentedItemsByShortName = completionProvider.getAllItemsByShortNameMatching(
             methodNameGetName,
-            (elem) => !!elem.documentation,
+            (elem) => elem.hasDocumentation(),
         );
 
         // should method documentation
```
mminot-yseop, line 62: What does that mean? xD

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -198,81 +156,22 @@ describe('Extension Server Tests', () => {
...
+            assertHasNelementsRemaining(
+                completionProvider.completions,
+                1,
+                (elem) => elem.kind === CompletionItemKind.Property,
+            );
+            assertHasNelementsRemaining(
+                completionProvider.completions,
+                2,
+                (elem) => elem.kind === CompletionItemKind.Method,
+            );
```
mminot-yseop, line 174: Kinda scary to see that the tests are shorter xD Do you check thinks as thoroughly as before? Is it just some kind of refacto?
Omg the old version looks dirty, actually xD

server/src/test/YmlKaoFileVisitor.test.ts
```diff
@@ -325,197 +224,27 @@ describe('Extension Server Tests', () => {
...
+                completionProvider.completions,
+                4,
+                (elem) => elem.kind === CompletionItemKind.Function,
+            );
             done();
         });
     });
 });
+
+function assertHasNelementsRemaining(
```
mminot-yseop, line 244: LOL I was wondering about the weird capitalization (`Nelements` vs. `NElements`). Didn’t realize that was something you wrote.
Does your linter yell at you if you use `…NE…`?

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -32,6 +32,10 @@ export abstract class AbstractYmlObject implements CompletionItem {
     public data?: any;
     /* End of overriden properties. */
 
+    private sourceElementName: string;
```
mminot-yseop, line 35: Not sure what this is.

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -32,6 +32,10 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
+    private documentationAsText: string;
+    public kindName: string = 'Object';
```
mminot-yseop, line 37: Not sure what this is either.

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -82,26 +86,58 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
             this.scopeEndOffset = scopeEndOffset;
             this.scopeStartOffset = scopeStartOffset;
         }
     }
 
+    public setDetail(type: string): void {
+        this.detail =
+            this.sourceElementName === 'STATIC'
+                ? `(${this.kindName}) [${this.sourceElementName}] ${this.label} ⇒ ${type}`
+                : `(${this.kindName}) [${this.sourceElementName}].${this.label} ⇒ ${type}`;
```
mminot-yseop, line 104: I _literally_ took me more than ten seconds to spot the difference, man.

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -82,26 +86,58 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
+    }
+
+    public getHoverContent(): MarkupContent {
```
mminot-yseop, line 107: It’s a (tiny) pity this is recomputed each time, but I dunno what’s best between this and having to update it in every relevant setter. :thinking: 

```java
private foo
private bar
private fooBar

constructor(…) {
  …
  updateFooBar()
}

getFooBar() {
  return fooBar
}

setFoo(x) {
  foo = x
  updateFooBar()
}

setBar(x) {
  bar = x
  updateFooBar()
}

private updateFooBar() {
  fooBar = foo + bar
}
```

LOL I’m soooo overthinking this!

server/src/yml-objects/YmlEnumMember.ts
```diff
@@ -9,6 +9,11 @@ export class YmlEnumMember extends AbstractYmlObject {
 
     constructor(indentifier: string, uri: string) {
         super(indentifier, CompletionItemKind.EnumMember, uri);
+        this.kindName = 'enum member';
+    }
+
+    public setDetail(type: string) {
+        return `(${this.kindName}) ${this.getShortName} ⇒ ${type}`;
```
mminot-yseop, line 16: Weird that you follow more or less the same pattern for every kind of symbol (“(foo) blahblah ⇒ plop”) but have to hardcode it in every class.

server/src/yml-objects/YmlObjectInstance.ts
```diff
@@ -8,7 +8,14 @@ export class YmlObjectInstance extends AbstractYmlObject {
...
     public readonly kind = CompletionItemKind.Variable;
 
-    constructor(indentifier: string, uri: string) {
+    constructor(indentifier: string, uri: string, public readonly isLocal) {
         super(indentifier, CompletionItemKind.Variable, uri);
+        this.kindName = isLocal ? 'variable' : 'instance';
+    }
+
+    public setDetail(type: string): void {
+        this.detail = this.isLocal
```
mminot-yseop, line 17: Just one `=`? :thinking: And weird comparison if it is one (details compared to being local?).

server/src/test/YmlCompletionItemsProvider.test.ts
alacambre-yseop: :rofl:

server/src/yml-objects/YmlObjectInstance.ts
alacambre-yseop: We create `YmlObjectInstance` when we encounter a static instance, a local variable or a function argument.
Here I make the distinction between static instances (not local) by adding `[GLOBAL]` to the detail and the rest.

server/src/yml-objects/AbstractYmlObject.ts
alacambre-yseop: Sorry :sob:

server/src/yml-objects/AbstractYmlObject.ts
alacambre-yseop: Well, in fact, once I'm sure I got both documentation and detail, I can save this as constant, since I recreate everything when there is a change in the file.

server/src/yml-objects/AbstractYmlObject.ts
alacambre-yseop: You're right. This misses documentation.

server/src/yml-objects/AbstractYmlObject.ts
alacambre-yseop: You're right. This misses documentation.

server/src/test/YmlKaoFileVisitor.test.ts
alacambre-yseop: I wasn't sure what to use. I think the linter won't yell.

server/src/test/YmlKaoFileVisitor.test.ts
alacambre-yseop: It's clearly less checked as before, but I think that trying to get full equality between objects was very hard to maintain.
Also, the main goal of these tests is mostly to check that we really get completion items and that they are ok-ish.

server/src/engineModel/EngineModel.ts
alacambre-yseop: Your version is better.

server/src/engineModel/EngineModel.ts
alacambre-yseop: * I do
* This method rewrite `detail`

> * This method rewrite `detail`

Oh, that's awful. :scream:

server/src/engineModel/EngineModel.ts
alacambre-yseop: `setUserInformations` sets both `detail` and `documentation`.

alacambre-yseop:
I have now a better implementation, I think.
I'm sorry if it seems messy. I had to add tests here and there to go above the 80% code coverage.

I added a base to finally test what happens when we parse a predefinedObject.xml file. :tada:

server/src/engineModel/EngineModel.ts
```diff
@@ -51,18 +51,19 @@ export class EngineModel {
...
-            ymlAbstractFunction.detail = `[${sourceType}].${ymlAbstractFunction.label}`;
+            ymlAbstractFunction.setUserInformations(
+                `(function) [${sourceType}].${ymlAbstractFunction.label}`,
+                documentation,
+            );
         } else {
-            ymlAbstractFunction = new YmlMethod(func.$.ident, this.uri);
-            ymlAbstractFunction.detail = `[STATIC] ${ymlAbstractFunction.label}`;
+            ymlAbstractFunction = new YmlFunction(func.$.ident, this.uri);
+            ymlAbstractFunction.setUserInformations(`(function) [STATIC] ${ymlAbstractFunction.label}`, documentation);
```
mminot-yseop, line 64: Strange that “function” is all quite while “STATIC” yells like crazy. (other occurrences elsewhere)

server/src/engineModel/EngineModel.ts
```diff
@@ -92,20 +93,17 @@ export class EngineModel {
...
         this.completionProvider.addCompletionItem(ymlClass);
     }
 
     private buildAttribute(attributeXmlElement: any, sourceType: string): YmlAttribute {
         const attribute = new YmlAttribute(attributeXmlElement.$.ident, this.uri);
-        attribute.detail = `[${sourceType}].${attribute.label}`;
-        if (!!attributeXmlElement.doc) {
-            attribute.setDocumentation(attributeXmlElement.doc[0]);
-        }
+        const documentation: string = !!attributeXmlElement.doc ? attributeXmlElement.doc[0] : null;
```
mminot-yseop, line 105: I’m at like 8 % of the diff and I saw that `!!foo.doc ? foo.doc[0] : null` pattern thrice already. :exploding_head:
What about making it more readable and stuff via a function? You may even already have declared one for that somewhere, I dunno.

server/src/visitors/YmlVisitorHelper.ts
```diff
@@ -0,0 +1,62 @@
+import { FieldContext } from '../grammar';
+import { connection } from '../server';
+
+const BEGINNING_QUOTES_REGEX = /^("""|")\s*/;
+const ENDING_QUOTES_REGEX = /\s*("""|")$/;
```
mminot-yseop, line 5: I had almost forgotten xd

```suggestion

/** Used to trim YML documentation. */
const BEGINNING_QUOTES_REGEX = /^("""|")\s*/;
/** Used to trim YML documentation. */
const ENDING_QUOTES_REGEX = /\s*("""|")$/;
```

server/src/visitors/YmlVisitorHelper.ts
```diff
@@ -0,0 +1,62 @@
...
+            connection.console.error(`${err}`);
+        } else {
+            connection.console.error(`An unexpected error occured when getting documentation value.`);
+        }
+    }
+    return null;
+}
+
+export function getType(fieldOptions: FieldContext[], baseType?: string): string {
+    let domains = baseType ? baseType : 'Object';
```
mminot-yseop, line 37: https://en.wikipedia.org/wiki/Elvis_operator

> TypeScript supports this concept with its nullish-coalescing operator ??, e.g. (a ?? b), since v3.7.[16]

LOL “nullish-coalescing” xD
Not sure if it is the same as falsy stuff, though. The empty strings seem to be kept, at least in the JS proposal https://github.com/tc39/proposal-nullish-coalescing. Grmbl.

server/src/yml-objects/YmlObjectInstance.ts
```diff
@@ -8,7 +8,14 @@ export class YmlObjectInstance extends AbstractYmlObject {
     public domainsLevel2: YmlType;
     public readonly kind = CompletionItemKind.Variable;
 
-    constructor(indentifier: string, uri: string) {
+    constructor(indentifier: string, uri: string, public readonly isLocal) {
```
mminot-yseop, line 11: What is that public readonly thingy? xD (I mean, those keywords)

alacambre-yseop: `public readonly` in a constructor allows to define and set the value of a final attribute for the corresponding class.
So, here as say that my class now has an `isLocal` final attribute and when the constructor is called, we set its value directly.

server/src/engineModel/EngineModel.ts
```diff
@@ -255,7 +257,33 @@ export class EngineModel {
...
+ * E.g. for the following XML:
+ * ```xml
+ * <xmlElement>
+ *     <doc>documentation</doc>
+ * </xmlElement>
+ * ```
+ * its representation is:
+ * ```JSON
+ * {
+ *     doc: ["documentation"]
```
mminot-yseop, line 279: * Not sure what you mean by “representation”. Is it what the function returns?
* The value seems to end up in a list, here (`[` `]`), but your doc says nothing of the sort (and the code seems to fetch `[0]`).

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -75,38 +84,69 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
      * @see MarkupContent
      */
-    public setDocumentation(doc: string): void {
-        if (!doc) {
-            this.documentation = null;
+    public setUserInformations(details: string, doc: string): void {
+        // Set documentation and detail attributes.
+        this.detail = details;
+        this.documentation = !!doc
+            ? {
```
mminot-yseop, line 129: LOL

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -75,38 +84,69 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
+                  value: doc,
+              }
+            : null;
+
+        if (!this.detail && !doc) {
+            return;
+        }
+
+        if (!this.detail) {
+            this.hoverContent = {
```
mminot-yseop, line 140: At this point, don’t we have the guarantee that `this.documentation` contains this `{ … }` thingy already? Could we reuse it? Hum not sure, dunno.
All these conditions make me dizzy. It would be great if you could spare a couple of minutes to try to make it simpler or clearer or at least add comments. I’m having a hard time figuring out which symbols can, at each point, be falsey or not, null or not, empty strings or not, etc.

server/src/engineModel/EngineModel.ts
alacambre-yseop: * We're using a package (`xml2js`) which transforms XML into a JSON representation. 
* Basically, this package creates a new object structure in which each XML child is an attribute named as the XML node and which value is an array. This array will contain the content of each node sharing the same name. The value can be an object or a string.

I'm not sure I'm very clear. Sorry.

mminot-yseop: Should the caller of the function give a damn about that internal JSON representation? Oh, you mean that we must pass as an argument a piece of XML that follows this representation. That could be explained in a better way, yes, I think xD

alacambre-yseop: Isn't the function documentation clear enough?
> Returns the value of the first child `doc` element within a given XML element, if any.

mminot-yseop: No, the documentation is misleading at best. You say “e.g.”, so I expect an example of input→output, and what you give instead are weird guidelines for how XML should be represented before calling the function.

I’d

1) replace “its representation is” with “it must be given as the following JSON representation, obtained thanks to xml2js”

2) add the corresponding function output, otherwise it’s not an example (or at least not one logically linked to the first sentence of the docs). Then the “e.g.” will make more sense.

alacambre-yseop: :+1:

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -75,37 +84,71 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
+
+        // If possible, set documentation's value.
+        if (!!doc) {
+            this.documentation = {
+                kind: MarkupKind.Markdown,
+                value: doc,
+            };
+        }
+
+        // There is no value for `details`. This means that we have a documentation.
```
mminot-yseop, line 138: The implication “no details ⇒ we have a documentation” is obscure and rather counter-intuitive to outsiders like me.

server/src/yml-objects/AbstractYmlObject.ts
```diff
@@ -75,37 +84,71 @@ export abstract class AbstractYmlObject implements CompletionItem {
...
+        // No need to set `this.detail`.
+        // Hover content is this.documentation.
+        if (!details) {
+            this.hoverContent = this.documentation;
+            return;
+        }
+
+        // At this point, we're not sur if we have a documentation,
+        // but we now that we have a value for `details`.
```
mminot-yseop, line 147:
```suggestion
        // At this point, we're not sure if we have documentation,
        // but we know that we have a value for `details`.
```

server/src/yml-objects/AbstractYmlObject.ts
alacambre-yseop: OK. Is it better?
```suggestion
        // There is no value for `details`. This means that we have a value for `doc`.
```

alacambre-yseop: Let me remind you that we have this code a few lines above:
```ts
if (!details && !doc) {
    return;
}
```

mminot-yseop: AH LOL I was thinking at a higher level of abstraction, like “there is some weird inner logic everywhere in the extension that ensures that when foo then bar”.

mminot-yseop: OK for your sugg

server/src/engineModel/EngineModel.ts
alacambre-yseop: Something like that?
image

mminot-yseop: Aye.

alacambre-yseop: Done.

PR #157, opened 2020-04-17
feat(grammar): improve parser

alacambre-yseop:
Multiple small changes with great impacts.
We go from 897 false positives to 89 false positives.

We now accept:
- `noDefault` keyword in switch expressions;
- accept `extends` as identifier;
- accept expressions like `myAttribute'` that are used in some constructions;
- accept ConstList objects as value for an attribute;
- add `switchExclusive` keyword;
- add a specific rule for time counters;
- accept key/value pairs list as attribute value;
- accept lists of constants as case in a switch instruction;

Still WIP: requires to add tests and update changelog.


alacambre-yseop:
You can still add comments and suggestions on this PR. I will ignore SonarCloud and force the merge since it detects duplication on code that is generated by ANTLR.
I have a task on my TODO list to tweak SonarCloud rules.

server/src/test/Parsing.test.ts
```diff
@@ -599,11 +648,35 @@ complete MyClass
...
     describe('ymlId rule', () => {
         it('should parse without errors different kinds of ymlId', (done) => {
-            checkInputValidityForRule((parser) => parser.ymlId(), `abc`);
-            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc`);
-            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc:d`);
-            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc::d`);
-            checkInputValidityForRule((parser) => parser.ymlId(), `a:bc:::d`);
+            const keywords = [
+                'args',
+                'local',
```
davanessoff-yseop, line 653: Do you have `static` defined somewhere?

alacambre-yseop: I add it.

mminot-yseop:
> We go from 897 false positives to 89 false positives.

Is that the number of reported errors when opening the reporting framework or something?

alacambre-yseop:
> > We go from 897 false positives to 89 false positives.
> 
> Is that the number of reported errors when opening the reporting framework or something?

Yes. Well, now it's 90 reported errors since the last commit. But it will be for later.

server/grammar/YmlLexer.g4
```diff
@@ -141,7 +142,7 @@ INTEGER: NON_ZERO_DIGIT DIGIT+ | DIGIT+;
  * − double colons are used to define a class' method;
  * − triple colons are used as a pointer towards a class' method or attribute.
  */
-YMLID: ID (COLON ID)* (COLON? COLON? COLON ID)?;
+YMLID: ID (COLON ID)* (COLON? COLON? COLON ID)? SINGLE_QUOTE?;
```
mminot-yseop, line 145: What the hell are single quotes doing at the end of IDs? xD Maybe you could update the fat comment above to explain it?

alacambre-yseop: I'm not even sure myself what this means. I know it can be used in a `new` instruction.
This instruction creates an object and can be used to set some of its attributes.
In javascript, you would have:

```javascript
{
    attr1: value1, 
    attr2: value2
}
```

In YML, assuming you're building an object of class MyClass,  it would be:
```yseopml
myObj = new(MyClass, attr1, value1, attr2, value2);
```

In case `attr1` and `attr2` are also identifiers of local variables, you can add a single quote to say that you are using the attribute name instead of the local variable.

```yseopml
function myFunction
args { }
local {
    Object attr1
    Object attr2
}
--> action {
// some code
myObj = new(MyClass, attr1', value1, attr2', value2);
// more code
};
```

mminot-yseop: o m g

mminot-yseop: I suppose asking you to edit the grammar so that this `'` would only be authorized in a `new(…)% would be overkill (and maybe even dangerous if you’re not sure whether there exist other valid uses).

PR #162, opened 2020-05-09
test: add tests for some classes; move code

Harukero:
Move code describing server capabilities to improve readability and future testing.
Tests for definition provider and parsing error listener

PR #163, opened 2020-05-11
test: error listener; definition provider

alacambre-yseop:
For an unknown reason the build for https://github.com/yseop/vscode-yseopml/pull/162 is not done. I cherry-picked the interesting commits and I will close that PR.

### DIFF
--- a/server/src/definitions/YmlDefinitionProvider.ts
+++ b/server/src/definitions/YmlDefinitionProvider.ts
@@ -2,7 +2,14 @@ import { Definition, Location } from 'vscode-languageserver';
 
 import { AbstractYmlObject } from '../yml-objects';
 
-export const LOCATION_FILTER = (uri: string) => (defLoc: AbstractYmlObject) => defLoc.definitionLocation.uri !== uri;
+/**
+ * Creates a filtering function over an AbstractYmlObject that will check that its definition location's uri
+ * is the same as the provided uri.
+ * @param uri the uri used to filter
+ */
+function createLocationFilter(uri: string) {
+    return (defLoc: AbstractYmlObject) => defLoc.definitionLocation.uri !== uri;
+}
 
 /**
  * Contains and provide entities definitions.
@@ -112,7 +119,7 @@ export class YmlDefinitionProvider {
      * @param uri The file's URI.
      */
     public removeDocumentDefinitions(uri: string): void {
-        this.definitions = this.definitions.filter(LOCATION_FILTER(uri));
+        this.definitions = this.definitions.filter(createLocationFilter(uri));
     }
 
     /**
@@ -120,6 +127,6 @@ export class YmlDefinitionProvider {
      * @param uri The file's URI.
      */
     public removeDocumentImplementations(uri: string): void {
-        this.implementations = this.implementations.filter(LOCATION_FILTER(uri));
+        this.implementations = this.implementations.filter(createLocationFilter(uri));
     }
 }
--- a/server/src/definitions/YmlDefinitionProvider.ts
+++ b/server/src/definitions/YmlDefinitionProvider.ts
@@ -64,7 +64,7 @@ export class YmlDefinitionProvider {
     }
 
     /**
-     * Add a definition to this provider. The element must have a definition location set.
+     * Add a definition to this provider. Elements without a set definition location will be ignored.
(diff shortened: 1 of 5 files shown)


server/src/definitions/YmlDefinitionProvider.ts
```diff
@@ -62,34 +64,62 @@ export class YmlDefinitionProvider {
     }
 
     /**
-     * Add a definition to this provider.
+     * Add a definition to this provider. The element must have a definition location set.
```
alacambre-yseop, line 67: How about this? @mminot-yseop 
```suggestion
     * Add a definition to this provider. Elements without a definition location set will be ignored.
```

mminot-yseop: > Elements without a definition location set will be ignored.

↓

> Elements without a set definition location will be ignored.

(“set” is both a noun and a verb, so let’s be careful xd I actually checked the code to see if it was a set, lol)

:heavy_plus_sign: do that for the other occurrence below.

alacambre-yseop:
```suggestion
     * Add a definition to this provider. Elements without a set definition location will be ignored.
```

server/src/definitions/YmlDefinitionProvider.ts
```diff
@@ -62,34 +64,62 @@ export class YmlDefinitionProvider {
...
     public addDefinition(def: AbstractYmlObject): void {
+        if (!def || !def.definitionLocation) {
+            return;
+        }
         this.definitions.push(def);
     }
 
     /**
-     * Add an implementation to this provider.
+     * Add an implementation to this provider. The element must have a definition location set.
```
alacambre-yseop, line 78:
```suggestion
     * Add an implementation to this provider. Elements without a set definition location will be ignored.
```

mminot-yseop:
Now maybe you can explain what this does and changes for the average user xD

alacambre-yseop:
> Now maybe you can explain what this does and changes for the average user xD

Sure, I just added some tests. It happened that I encountered some issues when no definition location was set, so I made it less error-prone.
It adds absolutely nothing for the end-user.

server/src/definitions/YmlDefinitionProvider.ts
```diff
@@ -2,12 +2,14 @@ import { Definition, Location } from 'vscode-languageserver';
 
 import { AbstractYmlObject } from '../yml-objects';
 
+export const LOCATION_FILTER = (uri: string) => (defLoc: AbstractYmlObject) => defLoc.definitionLocation.uri !== uri;
```
mminot-yseop, line 5: Not trivial to guess that this _excludes_ the items that bear the given URI. Docs? Maybe also longer name but not sure about this.

server/src/test/YmlParsingErrorListener.test.ts
```diff
@@ -0,0 +1,29 @@
...
+import { YmlParsingErrorListener } from '../visitors';
+
+function getDiagnostics(text: string): Diagnostic[] {
+    const diagnostics: Diagnostic[] = [];
+    const inputStream = CharStreams.fromString(text, '/path/to/file');
+    const lexer = new YmlLexer(inputStream);
+    const tokenStream = new CommonTokenStream(lexer);
+    const parser = new YmlParser(tokenStream);
+    parser.removeErrorListeners();
+    parser.addErrorListener(new YmlParsingErrorListener(diagnostics));
```
mminot-yseop, line 14: Does that constructor add stuff into `diagnostics`?

alacambre-yseop: Yes, it does. That's what is tested below.

alacambre-yseop: Ho. Sorry. The constructor doesn't add stuff into `diagnostics` but when an error is found, the `listener` adds stuff to `diagnostics`

mminot-yseop: Hum. Because when one reads that function too fast, they may wonder why we carry an empty array around.

alacambre-yseop: I added a comment. See 64c9f9b.

PR #165, opened 2020-06-02
fix(completion): stop sending full object on completion resolve request

alacambre-yseop:
There was a remaining issue when using the completion resolve request too many times. It happens to be due to the server sending full AbstractYmlObject objects. These objects are too heavy for the client when the list is too long.
Sending just the good amount of data (type, label, details and documentation) instead of the full objects solves the issue.

We'll be able to do a new release after this fix is merged.

### DIFF
--- a/server/src/features/CompletionResolveRequestHandler.ts
+++ b/server/src/features/CompletionResolveRequestHandler.ts
@@ -14,6 +14,15 @@ import { YmlCompletionItemsProvider } from '../completion/YmlCompletionItemsProv
  */
 export function completionResolveRequestHandler(completionProvider: YmlCompletionItemsProvider) {
     return (item: CompletionItem): CompletionItem => {
-        return completionProvider.completions.find((elem) => elem.data === item.data);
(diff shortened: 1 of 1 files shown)


server/src/features/CompletionResolveRequestHandler.ts
```diff
@@ -14,6 +14,15 @@ import { YmlCompletionItemsProvider } from '../completion/YmlCompletionItemsProv
  */
 export function completionResolveRequestHandler(completionProvider: YmlCompletionItemsProvider) {
     return (item: CompletionItem): CompletionItem => {
-        return completionProvider.completions.find((elem) => elem.data === item.data);
+        const candidate = completionProvider.completions.find((elem) => elem.data === item.data);
+        if (!candidate) {
+            return item;
+        }
+        // Set detail and documentation to `item` from the full AbstractYmlObject.
```
mminot-yseop, line 21: wut

```suggestion
        // Set the details and documentation of `item` according to those of the full AbstractYmlObject.
```

PR #168, opened 2020-07-10
feat(completion): handle inline named object instances

alacambre-yseop:
## Reminder
Files that are in `server/src/grammar/` are automatically generated from the grammar (`*.g4` files).


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## X.X.X (YYYY-MM-DD)
+
+-   Inline named object instances are now available when using completion lists, hover and other features like object instances are.
```
mminot-yseop, line 5: Lots of changes, so take this with a grain of salt.

```suggestion
-   Named inline object instances now appear (like “normal” instances already do) in lists provided for autocompletion, via hover, etc.
```

ahelman-yseop, line 5: j'ai pas compris le "are' a la fin...

server/src/visitors/YmlObjectInstanceVisitor.ts
```diff
@@ -25,5 +25,6 @@ export class YmlObjectInstanceVisitor extends YmlBaseVisitor {
         this.completionProvider.addCompletionItem(instance);
         instance.setDefinitionLocation(node.start, node.stop, this.uri);
         this.definitions.addDefinition(instance);
+        this.visitChildren(node);
```
alacambre-yseop, line 28: Required to visit the children that may be other inline instances.

server/grammar/YmlParser.g4
```diff
@@ -300,8 +300,10 @@ chainedCall:
     )*
 ;
 
+unnamedStaticDeclaration: className=ymlId (fieldValue)* SEMICOLON;
```
mminot-yseop, line 303: The fact that the new rule is for “unnamed” declarations makes me wonder if I understood correctly the RN entry above.

Is it “Inline-named (object instances)”, “Inline (named object instances)”? Or something else entirely? :exploding_head:

CHANGELOG.md
mminot-yseop: “de la même manière que les blabla le sont [déjà]”

mminot-yseop: I may need to better understand what those instances are to improve the suggestion.

server/grammar/YmlParser.g4
alacambre-yseop: It is `Inline (named object instances)`. I can probably rename this rule or just integrate it directly in `inlineDeclaration`, since it will never be used elsewhere?

CHANGELOG.md
mminot-yseop: (cf. comment in `YmlParser.g4`)

server/grammar/YmlParser.g4
mminot-yseop: Hum my question was, like, is it

named inline
vs.
unnamed inline

or

named inline
vs.
named not inline

I think your original RN entry hints towards the latter, while the code here does the opposite.

Pfiou, I think I finally succeeded in putting into words what’s been nagging me since the beginning.

mminot-yseop: (and regardless: no I think keeping separate rules, kinda like with functions, will make things clearer.)

mminot-yseop: oh but you removed it already lol. Do as you wish for that, I guess.

alacambre-yseop: We try in this PR to make public the instances defined as inline and having a name. So, we're more on a

inline and named
vs.
inline but unnamed

alacambre-yseop: I removed the new rule, anyway. So this won't be an issue anymore.

mminot-yseop: I still needed that answer to make sure my sugg on the RN wasn’t nonsensical. Seems OK. I’ll let the others test, though.

CHANGELOG.md
ahelman-yseop: aaaah puree je voyais ca "Inline named object instances are now available when using completion lists," puis une autre partie de phrase "hover and other features like object instances are." du coup ca n'avait aucn sens :D

ahelman-yseop: je me demande si on ne pouvait pas rajouter une virgule avant "like" pour que ca soit plus evident. mais pas sure... enfin bref c'est comprehensible mtn de toute facon ^^

PR #172, opened 2020-07-24
feat(format): new document format feature; improve settings system

alacambre-yseop:
Here is the new feature of document format ([VSC-4](https://yseop-knowledge.atlassian.net/browse/VSC-4)). It is still in Beta version but I think that it sets some good bases for future developments.

I made some changes in how we handle the settings. Before, we had to keep track of a lot of attributes. Now, I use a real object. It is clearer and makes easier future additions of options. ([VSC-5](https://yseop-knowledge.atlassian.net/browse/VSC-5)).

Also, you'll see in the code that I added some weird `yes`/`no` possible values to an option instead of a boolean.
I did this because it allows displaying the document format options as a table in the UI. If I don't do that, one will have to directly edit settings as JSON, which is less user-friendly.

Last but not least, I added the first _real_ client test. :tada: 

I had some difficulties on code coverage-side because of the settings change, which blocked me a little on adding other devs options. I'll do it when the main server file will be more covered.


server/src/server.ts
```diff
@@ -425,6 +407,49 @@ connection.onCompletion((pos: CompletionParams): CompletionItem[] => {
...
+});
+
+/**
+ * Build and send the list of Text Edit to apply to `document`
+ * accordingly to the settings defined in `documentFormatSettings`.
+ *
+ * @param document the text document to format
+ * @param documentFormatSettings the document format settings to apply
+ */
+export function buildDocumentEditList(document: TextDocument, documentFormatSettings?: IDocumentFormatSettings) {
```
alacambre-yseop, line 423: I first tried to put this code in another `feature` file but it failed because of some obscure cross dependency issues during the texts. No problem when this code is here. :shrug:

server/src/server.ts
```diff
@@ -425,6 +407,49 @@ connection.onCompletion((pos: CompletionParams): CompletionItem[] => {
     return completionProvider.getAvailableCompletionItems(pos.textDocument.uri, doc.offsetAt(pos.position));
 });
 
+connection.onDocumentFormatting((_params) => {
+    const doc = documents.get(_params.textDocument.uri);
+    // We should directly send an empty list if this feature is disabled by the settings.
```
alacambre-yseop, line 412: I struggled a lot with code coverage. I'm not sure if adding this condition now will make code coverage OK.
Just in case, I prefer to let the code like this and add a condition later.

server/src/settings/Settings.ts
```diff
@@ -0,0 +1,62 @@
...
+    /** `yes` to enable the document format feature. `no` otherwise. */
+    enableDocumentFormat: 'yes' | 'no';
+    /** `true` to add semicolon when possible. `false` otherwise. */
+    semicolonWhenPossible: boolean;
+    /**
+     * `true` to set a one space distance between some keywords and parenthesis. \
+     * `false` to have this distance set to zero.
+     */
+    spaceBetweenKeywordAndParenthesis: boolean;
+    // More options to come.
```
alacambre-yseop, line 30: At least, I'll have to add an option for the feature that set one space around `&&` `||` and such elements.

alacambre-yseop:
> Seems ok, but don't understand half of it.

Feel free to ask. I will most likely set Core team members as reviewers more and more.

mminot-yseop:
First and foremost (always hard to come back from vacations), the title is ambiguous, lol.
– doc sounds like documentation but seems to be, here, “document”.
– is “setting system” a “I set a system up” or “I added a system allowing VSC users to set things up”?

Plus, you say “beta” but it will probably be released just like the rest.

alacambre-yseop:
> First and foremost (always hard to come back from vacations), the title is ambiguous, lol.
– doc sounds like documentation but seems to be, here, “document”.
– is “setting system” a “I set a system up” or “I added a system allowing VSC users to set things up”?

I updated the title. This PR contains the new document format feature _and_ some internal improvements on how we handle the settings.

> Plus, you say “beta” but it will probably be released just like the rest.

It will be released, but this feature is really in “beta” because I suspect some potential issues. Since we rely on the ANTLR parser to format the document, a rule producing false positive may introduce some issues in the code and make it unable to be compiled.
Also, I think we'll have to rework a little some of the newly introduced settings.

server/src/server.ts
```diff
@@ -425,6 +407,49 @@ connection.onCompletion((pos: CompletionParams): CompletionItem[] => {
...
+    const tokenStream = new CommonTokenStream(lexer);
+    const parser = new YmlParser(tokenStream);
+    // No need to syntax errors to the client.
+    parser.removeErrorListeners();
+
+    // Parse the input.
+    const result = parser.kaoFile();
+    const edits: TextEdit[] = [];
+    const visitor = new YmlKaoFileVisitor(
+        new YmlCompletionItemsProvider(),
```
mminot-yseop, line 435: Useless here? Could be replaced with an idle or dummy object or even null? Not sure how that works.

server/src/server.ts
```diff
@@ -425,6 +407,49 @@ connection.onCompletion((pos: CompletionParams): CompletionItem[] => {
...
+        document.uri,
+        new YmlDefinitionProvider(),
```
mminot-yseop, line 437: Idem? Useless here? Could be replaced with an idle or dummy object or even null? Not sure how that works. Also dunno how costly it is to create this. But gosh we do so for every formatted file?

server/src/settings/Settings.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Default values for the document format settings, as defined in `package.json`.
```
mminot-yseop, line 34: So they _have to_ be defined in both places? :o

server/src/settings/Settings.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export const DEFAULT_DOC_FORMAT_SETTINGS: IDocumentFormatSettings = {
+    enableDocumentFormat: 'yes',
+    semicolonWhenPossible: true,
+    spaceBetweenKeywordAndParenthesis: false,
+};
+
+/**
+ * Set in an instance of document format settings set the default values for every undefined options.
```
mminot-yseop, line 43: Double “set”. + kinda the same confusion as in the AA IHM. (set a default value ≠ set a parameter to its default value)

```suggestion
 * In an instance of document format settings, set every undefined option to its default value.
```

server/src/settings/Settings.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @param documentFormat a document format settings set
```
mminot-yseop, line 45: Oh so some of those weird extra “set” are intentional? Still, not sure it’s necessary.

```suggestion
 * @param documentFormat document format settings, possibly with unset properties.
```

server/src/settings/Settings.ts
```diff
@@ -0,0 +1,62 @@
...
+ *
+ * @return the provided instance with default values for every attribute that was undefined
+ * or `DEFAULT_DOC_FORMAT_SETTINGS` if the instance was falsy.
```
mminot-yseop, line 48: > `DEFAULT_DOC_FORMAT_SETTINGS` if the instance was falsy.

Does this actually serve a purpose, or is it just a lame way to prevent not-yet-understood NPEs or something?

server/src/settings/Settings.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (!documentFormat) {
+        return DEFAULT_DOC_FORMAT_SETTINGS;
+    }
+    documentFormat.enableDocumentFormat =
+        documentFormat.enableDocumentFormat ?? DEFAULT_DOC_FORMAT_SETTINGS.enableDocumentFormat;
+    documentFormat.semicolonWhenPossible =
+        documentFormat.semicolonWhenPossible ?? DEFAULT_DOC_FORMAT_SETTINGS.semicolonWhenPossible;
+    documentFormat.spaceBetweenKeywordAndParenthesis =
+        documentFormat.spaceBetweenKeywordAndParenthesis ??
+        DEFAULT_DOC_FORMAT_SETTINGS.spaceBetweenKeywordAndParenthesis;
```
mminot-yseop, line 60: * Maybe you could do something similar to my…
https://github.com/PrismJS/prism/blob/master/components/prism-bash.js#L175-L195
… in order to just loop through properties (maybe even getting the list of relevant properties automatically, with no need for further maintenance when adding setting options)

* Huuuuuuuum won’t the `??` screw up if the option has a boolean type and was set to `false` by the user? If default=true and user-choice=false, won’t the user choice be overridden? Maybe I’m wrong.

server/src/test/DocumentFormatRequestHandler.test.ts
```diff
@@ -0,0 +1,150 @@
...
+    /* Two space additions + semicolon addition → “hashMap[country] = x;” */
+    hashMap[country]=x
+    /* Semicolon addition → “assert(a == b);” */
+    assert(a == b)
+    /* Semicolon addition → “return country;” */
+    return country
+};
+`,
+        );
+        expect(buildDocumentEditList(file, DEFAULT_DOC_FORMAT_SETTINGS)).toHaveLength(33);
```
mminot-yseop, line 86: I expected to find a comparison with the full resulting document :sweat_smile:

server/src/test/Settings.test.ts
```diff
@@ -0,0 +1,18 @@
+import {
+    DEFAULT_DOC_FORMAT_SETTINGS,
+    IDocumentFormatSettings,
+    setDocumentFormatDefaultValues,
+} from '../settings/Settings';
+
+describe('Settings Tests', () => {
+    it('should correctly set default format settings when there are no document format settings', (done) => {
```
mminot-yseop, line 8:
```suggestion
    it('should use default format settings when there are no document format settings', (done) => {
```

server/src/visitors/YmlBaseVisitor.ts
```diff
@@ -1,20 +1,242 @@
...
+        public document?: TextDocument,
     ) {
         super();
     }
 
     public defaultResult() {
         // no default
     }
+
+    public visitCombinedComparison(node: CombinedComparisonContext): void {
```
mminot-yseop, line 35: What are combined comparisons?

server/src/visitors/YmlBaseVisitor.ts
```diff
@@ -1,20 +1,242 @@
...
+        } else {
+            this.removeSpaceInterval(ifSymbolEnd, openParStart);
+        }
+
+        const closeParEnd = node.CLOSE_PAR().symbol.stopIndex;
+        const startIfBody = node.actionBlockOrInstruction()._start.startIndex;
+        sameLine = node.CLOSE_PAR().symbol.line === node.actionBlockOrInstruction()._start.line;
+        this.setOneSpaceInterval(closeParEnd, startIfBody, sameLine);
+        this.visitChildren(node);
+    }
```
mminot-yseop, line 165: Omg I thought most of the space-management stuff was built in and you just had to name a bunch of tokens, but you actually had to write heaps of boilerplate-ish stuff. :scream:

server/src/visitors/YmlBaseVisitor.ts
```diff
@@ -1,20 +1,242 @@
...
+            ),
+        );
+    }
+
+    /** Add an edit that adds a semicolon right after the provided node. */
+    private addSemiColon(node: ParserRuleContext): void {
+        this.filePossibleEdits.push(TextEdit.insert(this.document.positionAt(node.stop.stopIndex + 1), ';'));
+    }
+
+    /** Indicate if the document format is impossible and shouldn't be done. */
```
mminot-yseop, line 238: The docs and function name didn’t make me expect to find `docFormatSettings.enableDocumentFormat` here.

vscode-yseopml.code-workspace
```diff
@@ -10,7 +10,6 @@
             "formulahendry.auto-rename-tag",
             "formulahendry.auto-close-tag",
             "orta.vscode-jest",
-            "shinnn.stylelint",
```
mminot-yseop, line 13: ?

server/src/server.ts
```diff
@@ -425,6 +407,49 @@ connection.onCompletion((pos: CompletionParams): CompletionItem[] => {
...
+ * @param document the text document to format
+ * @param documentFormatSettings the document format settings to apply
+ */
+export function buildDocumentEditList(document: TextDocument, documentFormatSettings?: IDocumentFormatSettings) {
+    const inputStream = CharStreams.fromString(document.getText(), document.uri);
+    const lexer = new YmlLexer(inputStream);
+    const tokenStream = new CommonTokenStream(lexer);
+    const parser = new YmlParser(tokenStream);
+    // No need to syntax errors to the client.
+    parser.removeErrorListeners();
```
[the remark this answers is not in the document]
alacambre-yseop: > Oh but maybe that parser object is only used within this function anyway?

Yes. Each time we want to parse a file, we need to create a new Parser instance.
It looks weird so I'll try later a way to avoid multiple objects creation.

server/src/server.ts
alacambre-yseop: I'm not sure we have correct null-checks, so I prefer leaving it here.

server/src/server.ts
alacambre-yseop: I'm not sure we have correct null-checks, so I prefer leaving it here.
I'll make it cleaner later.

server/src/settings/Settings.ts
alacambre-yseop: Yes. The ones in `package.json` make them available through the UI. Here we define them programmatically.

server/src/settings/Settings.ts
alacambre-yseop: I prefer your suggestion. :+1:

server/src/settings/Settings.ts
alacambre-yseop: I prefer your suggestion. :+1:

server/src/settings/Settings.ts
alacambre-yseop: I'm not sure. :sweat_smile:

server/src/settings/Settings.ts
alacambre-yseop: > Maybe you could do something similar to my…
https://github.com/PrismJS/prism/blob/master/components/prism-bash.js#L175-L195
… in order to just loop through properties (maybe even getting the list of relevant properties automatically, with no need for further maintenance when adding setting options)

Maybe there is a better way to do so. https://attacomsian.com/blog/javascript-iterate-objects#:~:text=keys()%20Method-,The%20Object.,the%20value%20of%20each%20property.

> Huuuuuuuum won’t the `??` screw up if the option has a boolean type and was set to `false` by the user? If default=true and user-choice=false, won’t the user choice be overridden? Maybe I’m wrong.


`let x = foo ?? bar();` is equivalent to `let x = foo !== null && foo !== undefined ? foo : bar();`  (see [Typescript reference website](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing))

server/src/test/DocumentFormatRequestHandler.test.ts
alacambre-yseop: Couldn't do that. I wanted to do so but couldn't. :sob:

server/src/visitors/YmlBaseVisitor.ts
alacambre-yseop: `condition && condition`, or `condition || condition`

server/src/visitors/YmlBaseVisitor.ts
alacambre-yseop: Yeah. This was kinda boring. But now that it is done, I feel that adding an option will be a piece of cake.

server/src/visitors/YmlBaseVisitor.ts
alacambre-yseop: This doesn't bother me.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release Notes
 
+## X.X.X (YYYY-MM-DD)
+
+-   [BETA] New document formatting feature allowing small edits to your files. This feature can be deactivated using the Visual Studio Code settings.
```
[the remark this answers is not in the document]
mminot-yseop: Hum my sugg looks OK with regards to what you said there.

client/src/test/extension.test.ts
```diff
@@ -53,4 +32,25 @@ describe('Extension Tests', () => {
...
         });
     });
+    it('should get text edits from document format feature', (done) => {
+        const fileUri = vscode.workspace.textDocuments[0].uri;
+        expect(fileUri.path.endsWith('documentFormat.dcl')).toBeTruthy();
+        formatDocument(fileUri)
+            .then((list) => {
+                expect(list).toBeTruthy();
+                expect(list.length).toBe(8);
+            })
```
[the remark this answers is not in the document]
mminot-yseop: Only positions? No “insert _that_ text instead”, etc.?

If you can only rely on the cardinality of the change array, maybe it would be better to have concise examples instead of huge ones. When expecting 297529 changes, you end up having no idea where they come from; when expecting just a single change in a tiny example, it’s easier to avoid false negatives. But of course the best thing would probably be genuine “before vs. after vs. expectations” tests, with a full comparison to see if the formatter spouted exactly the document we expected.

client/src/test/extension.test.ts
```diff
@@ -53,4 +32,25 @@ describe('Extension Tests', () => {
...
+            .then(done, done);
+    });
 });
+
+/**
+ * Sends to the server a request to format the document at `docUri`.
+ *
+ * @param docUri uri to the document to format.
+ *
+ * @returns an array of text edits.
```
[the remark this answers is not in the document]
mminot-yseop: hum

package.json
```diff
@@ -89,6 +89,36 @@
...
+                    "type": "object",
+                    "default": {
+                        "enableDocumentFormat": "yes",
+                        "semicolonWhenPossible": true,
+                        "spaceBetweenKeywordAndParenthesis": false
+                    },
+                    "description": "Settings related to document format",
+                    "properties": {
+                        "enableDocumentFormat": {
+                            "type": "string",
```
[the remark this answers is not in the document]
mminot-yseop: Really? Booleans are not just checkboxes or something like this in the GUI version?
But then why stick to booleans for the semicolon thingy and the… other thing with a long name?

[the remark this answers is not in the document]
mminot-yseop: Note: Strings may allow further evolution, like “onlyIfBlahBlah”. Not sure it would be trivial to switch, on extension update, from a boolean to something like this xd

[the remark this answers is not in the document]
mminot-yseop: (I mean, two states vs. more than that)

server/src/visitors/YmlBaseVisitor.ts
mminot-yseop: But “condition” ≠ “comparison”, hence my confusion. This thing sounds awfully specific, as if it were only for “4 < 5” and not for “isPlop() == true”, for example.

server/src/visitors/YmlBaseVisitor.ts
mminot-yseop: I mean, “impossible” for something that the user delibirately disabled? Almost makes it sound like the user is struggling against the piece of software rather than using it.

alacambre-yseop: It's OK, we don't use `docFormatSettings.enableDocumentFormat` here anymore.

server/src/visitors/YmlBaseVisitor.ts
alacambre-yseop: I guess I'll have to take a deeper look at the name of the rules used in the grammar. I'll to it later.

package.json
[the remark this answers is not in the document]
alacambre-yseop: > Really? Booleans are not just checkboxes or something like this in the GUI version?

This is true only for first-level options. Here we are modifying the attributes of an object.

> But then why stick to booleans for the semicolon thingy and the… other thing with a long name?

I didn't like the yes-or-no thingy in the first place.

> Note: Strings may allow further evolution, like “onlyIfBlahBlah”.
(I mean, two states vs. more than that)

Exactly! I was thinking about it too.

> Not sure it would be trivial to switch, on extension update, from a boolean to something like this xd

I don't think so. We'll have to use deprecation stuff. :sweat:

client/src/test/extension.test.ts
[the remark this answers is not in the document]
alacambre-yseop: You're right. It may also increase the code coverage and test quality.

client/src/test/runTests.ts
```diff
@@ -15,7 +15,11 @@ async function main() {
         await runTests({
             extensionDevelopmentPath,
             extensionTestsPath,
-            launchArgs: ['--disable-extensions'],
+            launchArgs: [
+                `${extensionDevelopmentPath}/client/src/test/kb/`,
```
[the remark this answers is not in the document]
alacambre-yseop: Done. Please take a look.

package.json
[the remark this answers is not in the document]
mminot-yseop: I still don’t quite get why you don’t use the same thing for all three options, then xd
Sounds like since you don’t completely like yes/no, you only use it for one out of the three. But then you agree on the fact that yes/no is nicer for evolutions. :exploding_head:

client/resources/kb/_classes/Adress.yclass
```diff
@@ -0,0 +1,19 @@
+interface	Address
```
mminot-yseop, line 1: maybe store those files in a place that makes it clear they were poorly formatted on purpose.

client/src/test/runTests.ts
```diff
@@ -11,11 +11,15 @@ async function main() {
...
+        const resourcesDirPath = path.resolve(extensionDevelopmentPath, 'client/resources');
+        const kbDirPath = path.resolve(resourcesDirPath, 'kb');
+        const documentFormatFilePath = path.resolve(kbDirPath, '_technical/documentFormat.dcl');
+
         // Download VS Code, unzip it and run the integration test
         await runTests({
             extensionDevelopmentPath,
             extensionTestsPath,
-            launchArgs: ['--disable-extensions'],
+            launchArgs: [`${kbDirPath}/`, documentFormatFilePath, '--disable-extensions'],
```
mminot-yseop, line 22: * Not exactly what I had in mind xd
I asked if there could be two arguments instead of three, letting the test files themselves deduce `documentFormatFilePath` from `kbDirPath`. But maybe this way we’re more free (like, we can decorrelate the two paths in case `documentFormatFilePath` stops being under `kbDirPath`). :woman_shrugging: 

* Not sure why we’d need to append a slash to the directory path. That’s typically the type of things I try to avoid in Bash to make it simpler to concatenate stuff. I only append slashes for some logs, not in the actual variable values. But maybe here there’s some kind of obscure reason?

server/src/server.ts
```diff
@@ -82,6 +70,8 @@ const completionProvider: YmlCompletionItemsProvider = new YmlCompletionItemsPro
 // supports full document sync only
 const documents: TextDocuments = new TextDocuments();
 
+let SETTINGS: IYseopmlServerSettings;
```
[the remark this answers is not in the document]
mminot-yseop: > I think that even if it is not a `const` it feels better to have it with full caps since it is a global variable. I can rename it if required.

Not entirely convinced (the places where you assign values to it hurt my eyes xD) but not entirely against it.

server/src/test/DocumentFormatRequestHandler.test.ts
mminot-yseop: Then, as I said, it might be more useful and maintainable to split examples if you can. Parsing a huge chunk of code and then just checking that there were exactly 287562876 suggested changes sounds kinda weird. xD

mminot-yseop: If the test breaks because the number of changes went from 287562856 to 287562855, you may have a hard time finding out what exactly went wrong xd

client/src/test/runTests.ts
alacambre-yseop: The first arguments make VSCode open a directory or a file, depending on the end of the path. If the path ends with a slash, it is a directory (opening therefor some kind of workspace). If not, it will open or create a file, depending on whether the file exists or not.
I want to start my tests with a directory opened and a file open too.

package.json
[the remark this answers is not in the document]
alacambre-yseop: I was focused on UI instead of UI _and_ UX. I'll change it to yes/no everywhere.

server/src/server.ts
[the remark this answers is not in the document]
alacambre-yseop: I will rename it, then. At least to use the same standards as for the other global constants (such as `documents`).

server/src/test/DocumentFormatRequestHandler.test.ts
alacambre-yseop: You're right. I'll do the changes.

client/src/test/runTests.ts
mminot-yseop: Oh, crap. But that’s weird.

> If not, it will open or create a file, depending on whether the file exists or not.

Since it’s impossible to create a `foo` file when there’s already a `foo/` directory, a simple check would be able to make that “slash vs. no slash” distinction useless. Dunno what’s going on with that stuff.

client/resources/kb/_classes/Adress.yclass
mminot-yseop: Why is it poorly formatted, then? xDdd

mminot-yseop: Kinda scary to think that _this_ is representative of real projects xD

server/src/test/DocumentFormatRequestHandler.test.ts
mminot-yseop: Sounds rather good. I’ll try to take another look at the diff before too long.

server/src/server.ts
```diff
@@ -425,6 +407,51 @@ connection.onCompletion((pos: CompletionParams): CompletionItem[] => {
...
+        new YmlDefinitionProvider(),
+        documentFormatSettings,
+        edits,
+        document,
+    );
+
+    // Visit the result of the parsing.
+    // This fill the edits array.
+    visitor.visit(result);
+    if (parser.numberOfSyntaxErrors > 0) {
```
mminot-yseop, line 448: # Question

Can we do this check earlier, or is it `.visit(…)` that allows us to have this number of errors?

server/src/settings/Settings.ts
mminot-yseop: > `&&` `||` and such elements.

You mean “operators”? :trollface:
But seriously, I think you had grammar rules dedicated to conditions and boolean operators, so I,m genuinely wondering if you’ll be able to hadle all operators at once. Oh and some things like `.` or whatever may technically count as operators, in some contexts, I think :thinking: but obviously they do not follow the same spacing rules.

server/src/visitors/YmlBaseVisitor.ts
mminot-yseop: Did you take a proper note of https://github.com/yseop/vscode-yseopml/pull/172/files#r468649523?

mminot-yseop:
Aside from the few remaining questions, I tested and it seems OK.
Haha the file I opened at random had a `else{`. I hope you’ll soon have a feature to turn it into `else {` ^^

package.json
```diff
@@ -89,6 +89,44 @@
                     "default": "off",
                     "description": "Traces the communication between VSCode and the YML language server."
                 },
+                "yseopml.documentFormat": {
+                    "type": "object",
+                    "default": {
+                        "enableDocumentFormat": "yes",
+                        "semicolonWhenPossible": "yes",
+                        "spaceBetweenKeywordAndParenthesis": "no"
```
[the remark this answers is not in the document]
alacambre-yseop: Oh OK. But this is what we have in our YML libraries' code and what we set as default styling options in our `contributing.adoc` file.

server/src/server.ts
alacambre-yseop: I'm not sure. I'll do the check and change the code accordingly.

server/src/settings/Settings.ts
alacambre-yseop: “operators”, indeed.
In fact, I already handle − I think − all of those operators (`.` is not an operator in our case − or at least not defined as an operator).
I just don't have an option to remove this feature.

server/src/visitors/YmlBaseVisitor.ts
alacambre-yseop: Done here: https://yseop-knowledge.atlassian.net/browse/VSC-6

server/src/settings/Settings.ts
mminot-yseop: LOLed in the open space.

package.json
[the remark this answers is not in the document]
mminot-yseop: Kay, too bad xD Keep things as you wish, then, I guess.

server/src/server.ts
mminot-yseop: Kay.

alacambre-yseop: It was OK to avoid the `.visit(…)` call.

server/src/visitors/YmlBaseVisitor.ts
alacambre-yseop: In the end, I did a commit to fix this in the PR. :smile:

mminot-yseop (approved):

PR #177, opened 2020-11-24
feat: make functions and object instances foldable − VSC-8 VSC-9

alacambre-yseop:
In this PR I add a feature to make functions and object instances foldable.
image
There is an issue that has nothing to do with this PR where we don't save correctly the end of a function or method that has a multiline String as attribute's value.
This will be fixed in another PR.


erobert-yseop (dismissed):
Tests are ok, I only saw the issue you described with the multiline strings

alacambre-yseop:
:sob: 
Other blocks are not foldable anymore. It looks like the server's conf overrides what's defined client side.
I need to find a way to solve that first.

alacambre-yseop:
@erobert-yseop It should be better while still not perfect.
I managed to fix the issue we had with multiline Strings and have action blocks (blocks in functions that have instructions in it) foldable.
Adding more foldable things should be easy − depending on the feedbacks.

erobert-yseop:
@alacambre-yseop I tested and it seems great to me

server/grammar/YmlParser.g4
```diff
@@ -150,7 +150,7 @@ returnField:
 
 classPropertiesBlock: CLASSPROPERTIES classOption=field*;
 
-documentation: DOCUMENTATION;
+documentation: TRIPLE_QUOTE ANY* TRIPLE_QUOTE;
```
rsalmon-yseop, line 153: Is it possible to have a documentation with single quote string ? Or is it only for the fold so it is not relevant here ?

rsalmon-yseop (approved):
Can't review everything, but the grammar part looks ok.

server/grammar/YmlParser.g4
alacambre-yseop: Yes, that's possible. In fact, I should probably rename this rule to `multilineString`.
It's named like this because a long time ago I believed that multiline strings were accepted by YE only with this documentation attribute :sweat_smile:

server/src/features/FoldingRangeRequestHandler.ts
```diff
@@ -0,0 +1,29 @@
+import { FoldingRange, FoldingRangeParams } from 'vscode-languageserver';
+
+import { YmlDefinitionProvider } from '../definitions';
+import { AbstractYmlObject } from '../yml-objects';
+
+export function foldingRangeRequestHandler(definitionsProvider: YmlDefinitionProvider) {
```
mminot-yseop, line 6: Kinda funny that we say we’re about to handle a folding request or whatever, and only get weird YML definitions as an argument.
Actually, it seems that this function registers the ranges of all the things that could be folded, instead of trying to fold a particular item. Not what I expected from its name xd
Also curious (and related) that you went through the trouble of adding in-code comments but didn’t add a dedicated documentation for the function itself.

server/src/test/FoldingRangeRequestHandler.test.ts
```diff
@@ -0,0 +1,148 @@
...
+            FoldingRange.create(56, 56),
+            FoldingRange.create(57, 57),
+            FoldingRange.create(60, 76),
+            FoldingRange.create(62, 75),
+            FoldingRange.create(63, 66),
+            FoldingRange.create(66, 71),
+            FoldingRange.create(71, 74),
+            FoldingRange.create(78, 85),
+            FoldingRange.create(79, 81),
+            FoldingRange.create(82, 84),
```
mminot-yseop, line 137: xD just what I expected to see, but still kinda funny.

mminot-yseop (commented):
Just a poorly documented function.

server/src/features/FoldingRangesRequestHandler.ts
```diff
@@ -0,0 +1,40 @@
+import { FoldingRange, FoldingRangeParams } from 'vscode-languageserver';
+
+import { YmlDefinitionProvider } from '../definitions';
+import { AbstractYmlObject } from '../yml-objects';
+
+/**
+ * Create a request handler for the event `foldingRanges`.
```
alacambre-yseop, line 7: Is it better like this @mminot-yseop?

mminot-yseop: https://github.com/yseop/vscode-yseopml/pull/177#commitcomment-44732210 xd

mminot-yseop (approved):

PR #179, opened 2020-12-10
chore(release): version 1.12.0

alacambre-yseop:
Merge develop into master + update changelog and version number


erobert-yseop (approved):

mminot-yseop:
Seems to work.

Notes on folding: Kinda funny that I can fold these anonymous thingies:
image
… but not the `local` blocks:
image

mminot-yseop (approved):

alacambre-yseop:
> Seems to work.
> 
> Notes on folding: Kinda funny that I can fold these anonymous thingies:
> image
> … but not the `local` blocks:
> image

In the first example, we create the folding ranges by visiting the grammar rule `actionBlock`.
For the second example, we don't visit the rule that handles local blocks.
I'll do another PR for that during the month.
I'll add also the rules for collections too and stuff like that.

PR #187, opened 2021-01-28
fix: feedbacks from presentation + yseop deploy

alacambre-yseop:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## X.X.X (YYYY-MM-DD)
 
+-   Replaced command Yseop Info with Yseop Config.
+-   New command Yseop Libs which installs the libs required by your project for your OS.
+-   Fix a minor syntax highlighting issue with the `mod` operator.
+-   Replaced snippet `debug` with a new one named `logAttr` which is more explicit and which output is easier to read.
```
mminot-yseop, line 8: which outputs something
or
whose output is

```suggestion
-   Replaced the `debug` snippet with a new one, named `logAttr`, which is more explicit and whose output is easier to read.
```

vholten-yseop, line 8:
```suggestion
-   Replaced snippet `debug` with a new one named `logAttr` which is more explicit and whose output is easier to read.
```

server/src/features/CodeLensRequestHandler.ts
```diff
@@ -22,7 +22,10 @@ export function codeLensRequestHandler(definitionsProvider: YmlDefinitionProvide
             .map((func: YmlFunction) => {
                 const lens = CodeLens.create(func.definitionLocation.range);
                 lens.command = {
-                    title: `Cognitive complexity is ${func.getCognitiveComplexity()}.`,
+                    // Add a heart symbol when the complexity is equal or below 20.
```
vholten-yseop, line 25:
```suggestion
                    // Add a heart symbol when the complexity is equal to or below 20.
```

client/src/extension.ts
```diff
@@ -89,8 +89,12 @@ export function activate(context: ExtensionContext) {
...
     });
 
-    const infoCmd = commands.registerCommand(`${yseopmlSectionName}.info`, () => {
-        ExecYseopCliCommand(yseopCliPath, 'info');
+    const infoCmd = commands.registerCommand(`${yseopmlSectionName}.config`, () => {
+        ExecYseopCliCommand(yseopCliPath, 'config');
+    });
+
+    const libsInstallCmd = commands.registerCommand(`${yseopmlSectionName}.libs`, () => {
+        ExecYseopCliCommand(yseopCliPath, 'libs', 'install', '--os=auto', '-R');
```
mminot-yseop, line 97: If only one OS, the option-less `package` provided by the extension is almost doomed to fail.
I think we should try to make VSC-provided YCLI commands work together as much as humanely possible.

package.json
```diff
@@ -30,7 +30,8 @@
         "onCommand:yseopml.clean",
         "onCommand:yseopml.cleanall",
         "onCommand:yseopml.package",
-        "onCommand:yseopml.info"
+        "onCommand:yseopml.info",
```
mminot-yseop, line 33: still named `info` here?

server/src/features/CodeLensRequestHandler.ts
```diff
@@ -22,7 +22,10 @@ export function codeLensRequestHandler(definitionsProvider: YmlDefinitionProvide
...
+                    title: `${
+                        func.getCognitiveComplexity() <= 20 ? '$(heart) ' : ''
```
mminot-yseop, line 27: 20 seems kinda arbitrary and large xD

alacambre-yseop: In fact, according to [this post](https://community.sonarsource.com/t/how-to-use-cognitive-complexity/1894/4) it completely depends on the language.
For YML, I judged it high enough but not too high.

vholten-yseop (approved):

alacambre-yseop:
Build fails because of breaking changes on VSCode 1.53, producing issues with tests. :sob:

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## X.X.X (YYYY-MM-DD)
 
+-   Replaced command Yseop Info with Yseop Config.
+-   New command Yseop Libs which installs the libs required by your project.
```
mminot-yseop, line 6: Currently, there can be at most one, right?

```suggestion
-   New “Yseop Libs” command which installs the YML library required by your project.
```

server/src/test/CodeLensRequestHandler.test.ts
```diff
@@ -113,12 +113,16 @@ while(a > 0 // +1
         while(true == true) { // +3 (increment = 2)
             if(condition == true) { // +4 (increment = 3)
                 break; // +1
+            } else {
+                if(otherCondition == true) { // +5 (increment = 4)
+                    break; +1
```
mminot-yseop, line 118: ?

```suggestion
                    break; // +1
```

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## X.X.X (YYYY-MM-DD)
 
+-   Replaced command Yseop Info with Yseop Config.
+-   New “Yseop Libs” command which installs the YML library required by your project.
+-   New contextual menu entry on \*.yma file allowing to launch `yseop manager deploy` using Yseop CLI.
```
mminot-yseop, line 7:
```suggestion
-   New contextual menu entry on `*.yma` files allowing to launch `yseop manager deploy` using Yseop&nbsp;CLI.
```

client/src/extension.ts
```diff
@@ -89,8 +90,26 @@ export function activate(context: ExtensionContext) {
...
+        ExecYseopCliCommand(yseopCliPath, 'config');
+    });
+
+    const libsInstallCmd = commands.registerCommand(`${yseopmlSectionName}.libs`, () => {
+        ExecYseopCliCommand(yseopCliPath, 'libs', 'install', '-R');
+    });
+
+    const deployCmd = commands.registerCommand(`${yseopmlSectionName}.managerDeploy`, (ymaUri: Uri) => {
+        window
+            .showInputBox({ prompt: 'Leave this box empty to use the default env.', placeHolder: 'DEMO' })
```
mminot-yseop, line 103: * env → environment
* Having “DEMO” as an hardcoded placeholder seems dangerous to me. This gives the impression that it is the “default env” you’re talking about. Either make the default configurable and align the placeholder with it, or leave it empty, because the default env does not use any “key string”: it’s just `YSEOP_CLI_MANAGER_URL/USERNAME/PASSWORD` stuff.

client/src/extension.ts
```diff
@@ -89,8 +90,26 @@ export function activate(context: ExtensionContext) {
...
+            .then((env) => {
+                if (!!env && env.length > 0) {
+                    window.showInformationMessage(`Deploying YMA ${ymaUri.fsPath} on env “${env}”.`);
+                    ExecYseopCliCommand(yseopCliPath, 'manager', 'deploy', '-e', `${env}`, ymaUri.fsPath);
```
mminot-yseop, line 107: `-e ''` is accepted as an equivalent of “no -e at all”, because I do “if is not blank” and stuff. So your `if` seems a bit overkill.

client/src/extension.ts
```diff
@@ -89,8 +90,26 @@ export function activate(context: ExtensionContext) {
...
+    const libsInstallCmd = commands.registerCommand(`${yseopmlSectionName}.libs`, () => {
+        ExecYseopCliCommand(yseopCliPath, 'libs', 'install', '-R');
+    });
+
+    const deployCmd = commands.registerCommand(`${yseopmlSectionName}.managerDeploy`, (ymaUri: Uri) => {
+        window
+            .showInputBox({ prompt: 'Leave this box empty to use the default env.', placeHolder: 'DEMO' })
+            .then((env) => {
+                if (!!env && env.length > 0) {
+                    window.showInformationMessage(`Deploying YMA ${ymaUri.fsPath} on env “${env}”.`);
```
mminot-yseop, line 106: * “on”? 🤔 why not “with”, like in the case below?
* as before, “environment” seems friendlier than “env”. Even the option’s long name is “environment”; let’s try to align a bit and avoid unnecessary abbreviations.

client/src/extension.ts
```diff
@@ -89,8 +90,26 @@ export function activate(context: ExtensionContext) {
...
+                    ExecYseopCliCommand(yseopCliPath, 'manager', 'deploy', '-e', `${env}`, ymaUri.fsPath);
+                } else {
+                    window.showInformationMessage(`Deploying YMA ${ymaUri.fsPath} with default env.`);
+                    ExecYseopCliCommand(yseopCliPath, 'manager', 'deploy', ymaUri.fsPath);
+                }
```
mminot-yseop, line 111: Maybe add one -v, if the output appears in a log thingy?

client/src/extension.ts
alacambre-yseop: OK!

client/src/extension.ts
alacambre-yseop: OK!

client/src/extension.ts
alacambre-yseop: > So your if seems a bit overkill.

I prefer to have a case where we have a nice message saying that we're using the default environment.
I'll let the `if` statement as it is.

client/src/extension.ts
alacambre-yseop: OK!

client/src/extension.ts
mminot-yseop: For the message, sure, but then why duplicate the command? You could do

```
if …
  message
else
  other message

command
```

or use a ternary operator or whatever. But having two `ExecYseopCliCommand` here seems useless. That was my point.

alacambre-yseop: Alright!
Done.

mminot-yseop (approved):

PR #191, opened 2021-02-17
feat: add project picker and progress bar

alacambre-yseop:
I'm having fun exploring the VSCode API.
I added a KB selector when a project contains multiple KB.
I also added a progress bar :)

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -5,6 +5,7 @@
 'use strict';
 import { exec } from 'child_process';
 import * as fs from 'fs';
+import glob from 'glob-promise';
 import * as path from 'path';
 import {
     commands,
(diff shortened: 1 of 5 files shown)


erobert-yseop (dismissed):
Tests of KB selector and progress bar are OK :+1:

client/src/extension.ts
```diff
@@ -169,9 +172,41 @@ async function ExecYseopCliCommandWithKbDirectoryPath(cliPath: string, ...words:
         window.showErrorMessage('This command must be used from within a workspace folder.');
         return;
     }
-
-    const workspaceFolders = workspace.workspaceFolders;
-    ExecYseopCliCommand(cliPath, ...words, workspaceFolders[0].uri.fsPath);
+    const workspaceFirstDir = workspace.workspaceFolders[0].uri.fsPath;
+
+    // Get the list of possible KBs in the workspace (like in library projects)
```
mminot-yseop, line 177: Write proper functions and this comment will become virtually useless xD

client/src/extension.ts
```diff
@@ -169,9 +172,41 @@ async function ExecYseopCliCommandWithKbDirectoryPath(cliPath: string, ...words:
...
+    // using com.yseop.designer.prefs paths.
+    glob('**/.settings/com.yseop.designer.prefs', {
+        cwd: workspaceFirstDir,
+    })
+        .then((_matches) =>
+            // from “library/.settings/com.yseop.designer.prefs” generates “library”.
+            // from “.settings/com.yseop.designer.prefs” generates “.”
+            _matches.map((designerPrefsFile) => path.normalize(path.join(path.dirname(designerPrefsFile), '..'))),
```
mminot-yseop, line 185: Why not dirname² instead of this weird mix of dirname+..?

client/src/extension.ts
```diff
@@ -169,9 +172,41 @@ async function ExecYseopCliCommandWithKbDirectoryPath(cliPath: string, ...words:
...
+            qp.items = _matches.map((dirPath) => {
+                return {
+                    label: dirPath,
+                };
+            });
+            qp.onDidAccept(() => {
+                // When user hit Enter, hide the quick picker
+                // and execute the CLI command using as kb path
+                // the result of a join of workspaceDir and the seleted item.
+                const selectedLabel = qp.selectedItems[0]?.label;
```
mminot-yseop, line 203: Doesn’t that `?.` mean that you may end up with a null label? What will happen then?

client/src/extension.ts
alacambre-yseop: That's not completely true, since it will be part of the documentation of the new function :p
It is clear that having a function dedicated to this part would make the code easier to read.
I'll do it.

client/src/extension.ts
alacambre-yseop: I just didn't think about it.
I'll try it and let you know.

client/src/extension.ts
alacambre-yseop: In fact, we will never get `null` at this point. I will remove the question mark.

mminot-yseop (approved):
If the testers are ok…

erobert-yseop (approved):
Tests are ok but is it normal that we don't get any output or progress bar when we use an Yseop command while we don't have the output window open? For example nothing happens if I call the command when I'm on the terminal or debug console

alacambre-yseop:
> Tests are ok but is it normal that we don't get any output or progress bar when we use an Yseop command while we don't have the output window open? For example nothing happens if I call the command when I'm on the terminal or debug console

Yeah, for a weird reason we are expecting an editor to be open in order to launch the command.
Removing this check seems having absolutely no effect.
Would you prefer that I remove it in this PR or in another?

erobert-yseop:
> > Tests are ok but is it normal that we don't get any output or progress bar when we use an Yseop command while we don't have the output window open? For example nothing happens if I call the command when I'm on the terminal or debug console
> 
> Yeah, for a weird reason we are expecting an editor to be open in order to launch the command.
> Removing this check seems having absolutely no effect.
> Would you prefer that I remove it in this PR or in another?

Well the issue comes from the handling of the command and not really the kb selector and progress bar so you can do this in another PR if you want

alacambre-yseop:
OK!

PR #192, opened 2021-02-19
feat: no need of an active text editor to launch commands

alacambre-yseop:
An open and active editor was required to launch a command.
This doesn't make sense so I removed this requirement.

### DIFF
--- a/client/src/extension.ts
+++ b/client/src/extension.ts
@@ -223,11 +223,6 @@ export async function ExecYseopCliCommand(cliPath: string, ...words: string[]) {
     yseopCliOutputChannel.clear();
     yseopCliStatusBarItem.hide();
 
-    const editor = window.activeTextEditor;
-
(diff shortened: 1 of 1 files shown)


client/src/extension.ts
```diff
@@ -223,11 +223,6 @@ export async function ExecYseopCliCommand(cliPath: string, ...words: string[]) {
     yseopCliOutputChannel.clear();
     yseopCliStatusBarItem.hide();
 
-    const editor = window.activeTextEditor;
-
-    if (editor === null || editor === undefined) {
-        return;
```
alacambre-yseop, line 229: Useless check. Also, no warning at all to tell the user. :shrug:

mminot-yseop (approved):
Hum, I hope we won’t suddenly discover that it was used in an hidden thingy.

erobert-yseop (approved):
That's better, thanks!

PR #194, opened 2021-03-16
feat(gotodef): go to definition on classnames

alacambre-yseop:
Allow using Go To Definition (shortcut F12) on class names.
Useful to check parent classes.

In the ticket I also wanted to take into account `complete` instructions, but it would require more devs than expected, so I won't do it here.
I need to think of a nice way to differentiate `complete MyObject;` from `complete MyClass;`, for instance.

I also wanted to make the Go To Definition point directly to predefinedObjects.xml file when a user has defined one, but the parser I use doesn't have any clue on the current line.
Sorry.

See https://yseop-knowledge.atlassian.net/browse/VSC-12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## X.X.X (YYYY-MM-DD)
 
+-   Go To Definition now works for class names too.
+
 ## 1.13.0 (2021-02-12)
 
 -   Replaced command Yseop Info with Yseop Config.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## X.X.X (YYYY-MM-DD)
 
--   Go To Definition now works for class names too.
+-   _Go To Definition_ now works for class names too.
 
 ## 1.13.0 (2021-02-12)
 
--- a/server/src/definitions/DefinitionUtils.ts
+++ b/server/src/definitions/DefinitionUtils.ts
@@ -93,5 +93,10 @@ export function getYmlEntityNamePartWithoutClassName(text: string, position: Ent
         return null;
     }
     const parts = result.split('::');
-    return parts[position === EntityPartPosition.END ? parts.length - 1 : 0];
+    if (position === EntityPartPosition.END) {
+        // We want to get the text from the end of the string.
+        return parts[parts.length - 1];
+    }
+    // We want to get the text from the start of the string.
+    return parts[0];
 }
--- a/server/src/definitions/DefinitionUtils.ts
+++ b/server/src/definitions/DefinitionUtils.ts
@@ -93,5 +93,5 @@ export function getYmlEntityNamePartWithoutClassName(text: string, position: Ent
         return null;
     }
     const parts = result.split('::');
-    return parts[parts.length - 1];
+    return parts[position === EntityPartPosition.END ? parts.length - 1 : 0];
 }
--- a/server/src/test/DefinitionUtils.test.ts
+++ b/server/src/test/DefinitionUtils.test.ts
@@ -117,10 +117,7 @@ describe('Extension Server Tests', () => {
             });
             it('should accept identifiers with namespaces', (done) => {
                 const source = 'Lang:Verb::VERB_JA_食べ物(';
-                expect(
-                    getYmlEntityNamePart(source, EntityPartPosition.BEGINNING)).toBe(
-                    'Lang:Verb::VERB_JA_食べ物',
-                );
+                expect(getYmlEntityNamePart(source, EntityPartPosition.BEGINNING)).toBe('Lang:Verb::VERB_JA_食べ物');
                 done();
             });
         });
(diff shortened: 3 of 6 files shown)


mminot-yseop:
We already had this on methods, right?

alacambre-yseop:
> We already had this on methods, right?

Yep. And also on object instances.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## X.X.X (YYYY-MM-DD)
 
+-   Go To Definition now works for class names too.
```
mminot-yseop, line 5: Was about to suggest using the UI macros, then remembered this was not Asciidoc. 😭
Still, some kind of emphasis on the feature name would make this feel more natural.

```suggestion
-   _Go To Definition_ now works for class names too.
```

server/src/definitions/DefinitionUtils.ts
```diff
@@ -93,5 +93,5 @@ export function getYmlEntityNamePartWithoutClassName(text: string, position: Ent
         return null;
     }
     const parts = result.split('::');
-    return parts[parts.length - 1];
+    return parts[position === EntityPartPosition.END ? parts.length - 1 : 0];
```
mminot-yseop, line 96: I… think I got part of this, but hard to be sure. What about a comment (and an update of the description of the parameter, to explain why `position` can now have this special-ish value, what it means and what will ensue)?
(Next step if I still don’t understand: ask for an example xD)

server/src/test/DefinitionUtils.test.ts
```diff
@@ -176,6 +173,23 @@ describe('Extension Server Tests', () => {
                 }
                 done();
             });
+
+            it('should find the class name when before double column, method name otherwise', (done) => {
```
mminot-yseop, line 177: Wait… Is that…?

```suggestion
            it('should find the class name when before double colon, method name otherwise', (done) => {
```

server/src/test/DefinitionUtils.test.ts
```diff
@@ -176,6 +173,23 @@ describe('Extension Server Tests', () => {
...
+                const source: string = 'mySuperClass::myMethod';
+                const leftPart = 'mySuperClass';
+                const leftPartSize = leftPart.length;
+                // From start to the end leftPart
+                for (let position = 0; position < leftPartSize + 1; position++) {
+                    expect(getTokenAtPosInDoc(source, position)).toBe(leftPart);
+                }
+                // Between the two colons
+                expect(getTokenAtPosInDoc(source, leftPartSize + 1)).toBe(null);
```
mminot-yseop, line 186: LOL this is not mentioned by the test’s description.

server/src/test/FoldingRangesRequestHandler.test.ts
```diff
@@ -116,6 +116,7 @@ Country myCountry // 78 → 85
         const foldingRanges: FoldingRange[] = handler(DEFAULT_REQUEST_PARAMS);
         expect(foldingRanges).toBeTruthy();
         const expectedFoldingRanges = [
+            FoldingRange.create(1, 1), // Class name
```
mminot-yseop, line 119: Noob question: Does 1–1 create, like, a visible folding button that does nothing?

server/src/visitors/YmlEnumVisitor.ts
```diff
@@ -31,6 +31,9 @@ export class YmlEnumVisitor extends YmlBaseVisitor {
         this.yenum.enrichWith(doc, this.enumName);
         this.completionProvider.addCompletionItem(this.yenum);
         this.yenum.setDefinitionLocation(node.start, node.stop, this.uri);
+        // Add enum to the definitions, because it looks a lot like a class.
+        this.definitions.addDefinition(this.yenum);
+        // Keep this for compatibility.
```
mminot-yseop, line 36: Compatibility of what with what?

mminot-yseop:
Test seems OK (jumped from implem to declaration of a dumb GeneralData I found).

erobert-yseop (dismissed):
Tests are OK

server/src/visitors/YmlEnumVisitor.ts
alacambre-yseop: _Go To Implementation_ was already working for `enum`. But when using _Go To Definition_ on enums as for classes since it wasn't working it felt weird. That's why I know have both.

server/src/test/FoldingRangesRequestHandler.test.ts
alacambre-yseop: It does nothing.

server/src/visitors/YmlEnumVisitor.ts
mminot-yseop: I don’t understand that part: “when using Go To Definition on enums as for classes since it wasn't working”.
I also don’t get if the initial “Go To Implementation” is a typo meant to be “… Declaration” or if it’s meant to be “Implementation”.
All in all, I still can’t answer the question of “compatibility of what with what”. Basically, what happens if you remove the line that follows that comment.

alacambre-yseop: If I remove the line that follows, users won't be able to use _Go To Implementation_ for enums − if they were using it.

mminot-yseop: Then I’d simply say something like ‘For “Go to implementation”.’ xD dunno

mminot-yseop (approved):

PR #198, opened 2021-04-27
chore: update dependencies for lsp 7.0.0

alacambre-yseop:
Just some dependency upgrades.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -1,8 +1,11160 @@
 {
     "name": "vscode-yseopml",
     "version": "1.13.0",
-    "lockfileVersion": 1,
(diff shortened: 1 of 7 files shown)


mminot-yseop:
Probably not related to the bumps, but… `.fct` files like `configLibActors.fct` are detected as plain text? 🤔

mminot-yseop (approved):

alacambre-yseop:
> Probably not related to the bumps, but… `.fct` files like `configLibActors.fct` are detected as plain text?

Yes, `fct` is not in the set of extensions known to the extension.

alacambre-yseop:
Here are the file extensions we handle:
image

PR #199, opened 2021-04-27
feat: go to def to predefined objects XML file − VSC-15

alacambre-yseop:
The goal here is when using Go To Definition feature being able to directly go to the correct position in the user defined `predefinedObjects.xml` file.


ahelman-yseop (commented):
seems ok but i'll let someone who knows the language better approve

rea-yseop (approved):
Seems fine (only really looked the documentation though)

rea-yseop (approved):

PR #203, opened 2021-05-28
feat(variables): try to find unused variables and args

alacambre-yseop:
Adds an information/warning for function arguments or local variables that might be unused.
The message is not precise on purpose because the current state of the parser isn't enough to be 100% sure that the variable is unused.


ahelman-yseop (approved):
seems good :ok_hand:

PR #204, opened 2021-05-28
feat: improve parser

alacambre-yseop:
Makes parser way faster by making semicolon mandatory.
Also adds a lot of new parsing rules to handle more cases.
Also removed some ambiguities.


alacambre-yseop:
I you find examples that doesn't work when parsing, please open a ticket or give them as simplified as possible so that we improve the parser and integrate them in the tests.

obatier-yseop (approved):

ahelman-yseop (approved):

PR #209, opened 2021-06-28
fix: use dynamic document format settings instead of null one

alacambre-yseop:
We were using the value of configuration set at startup (which was null) instead of the dynamicaly computed configuration. The issue happened not only on the file sent by @obatier-yseop but in any other file too.

Maybe this PR code coverage won't pass because I had to do some changes on `server.ts`.I will merge it anyway ^^'.
Another PR (#205) will make code coverage way better once finished.

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -413,7 +413,7 @@ connection.listen();
 /**
  * Create a request handler for the event `documentFormatting`.
  * The handler will parse the current file and build and send the list of Text Edit
- * to apply client-side accordingly to the user settings.
+ * to apply client-side according to the user settings.
  *
  *
  * @param documents a manager for simple text documents
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -350,7 +350,15 @@ function parseFile(textDocUri: string, docContent: string) {
     // Reset all the document's completion items.
     completionProvider.removeDocumentCompletionItems(textDocUri);
 
-    const visitor = new YmlKaoFileVisitor(completionProvider, textDocUri, definitionsProvider, null, null, null, diagnostics);
+    const visitor = new YmlKaoFileVisitor(
+        completionProvider,
+        textDocUri,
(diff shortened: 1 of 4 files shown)


server/src/server.ts
```diff
@@ -395,3 +401,20 @@ connection.onCodeLens(codeLensRequestHandler(definitionsProvider));
 
 // Listen on the connection
 connection.listen();
+
+/**
+ * Create a request handler for the event `documentFormatting`.
+ * The handler, when receiving a `DocumentFormattingParams` instance will find
+ * will parse the current file and build and send the list of Text Edit
```
mminot-yseop, line 408: will find will parse? xD (plus missing comma).
Did you merge two existing sentences or something? xD

server/src/server.ts
```diff
@@ -404,9 +412,8 @@ connection.listen();
 
 /**
  * Create a request handler for the event `documentFormatting`.
- * The handler, when receiving a `DocumentFormattingParams` instance will find
- * will parse the current file and build and send the list of Text Edit
- * to apply client-side accordingly to the settings defined in `documentFormatSettings`.
+ * The handler will parse the current file and build and send the list of Text Edit
+ * to apply client-side accordingly to the user settings.
```
mminot-yseop, line 416:
```suggestion
 * to apply client-side according to the user settings.
```

mminot-yseop (approved):
Nothing looks too alarming. Kinda surprised to see a brand new handler but I trust you on that.
I’ll try to remember about the formatter during pre-release tests. I kinda forgot it even existed, TBH.

alacambre-yseop:
> Kinda surprised to see a brand new handler but I trust you on that.

I just moved the already existing handler so that it uses the configuration contained in `server.ts` that is dynamically updated.

This, in `server/src/features/DocumentFormattingRequestHandler.ts`:
image

Just became, in `server/src/server.ts`:
image

PR #215, opened 2022-01-25
feat: icons for yseopml files; .yml not yseopml

alacambre-yseop:
Please take some moments to test this.
Icons should be visible when the next version of VSCode (January 2022) is released (planned for February 03, 2022).
However, the extension should still work for older versions.


aauzimour-yseop (dismissed):
nice-looking feature !

erobert-yseop (dismissed):
So if I understand well, it's normal that I don't see the new icons
Otherwise I didn't see any issue :+1:

aauzimour-yseop:
Yes for now I couldn't see it as well. Will be visible in a month or two.

alacambre-yseop:
OK, so if we're lucky and I understand well, when you will update your version of VSCode you should see icons appear if you stick with the tested version of the extension.

alacambre-yseop:
> Icons should be visible when the next version of VSCode (January 2022) is released (planned for February 03, 2022).

mminot-yseop (changes requested):
SVG please. `fill="#blahblah"` in the `<path>` for yellow.

mminot-yseop (approved):
thx

